PR #200, opened 2014-08-13
Added control which fields can be used for ordering.

### LINKED ISSUE #199: Lack of control which fields used for ordering by Paginator
I think that's a big loss that there is no control on whitelist for Paginator:
http://book.cakephp.org/2.0/en/core-libraries/components/pagination.html#control-which-fields-used-for-ordering

redd:
Fixes #199.


jippi:
tests broken - way way too old

PR #229, opened 2014-12-02
Cake3 ApiListener Exception Fix

### LINKED ISSUE #228: Crud v4 ApiListener: ValidationException are not getting rendered
When a ValidationException is thrown this is the whole response:
```
{
    "message": "A validation error occurred",
    "url": "\/...\/965aceed-0f0c-4f04-906c-998fb9a60364.json",
    "code": 412
}
```
There are no details about the validation error. I think this due to the following facts:

1. Adding ApiListener to my project does not change the default ExceptionRenderer. I think the problem is registerExceptionHandler() in ApiListener is only changing the configure value for Error.exceptionRenderer. But the ErrorHandler was loaded in bootstrap.php so only the configurations in app.php are consumed.

2. When I change the ExceptionRender my app.php to Crud\Error\ExceptionRenderer I am getting the following error:
```
Call to a member function send() on a non-object in \vendor\cakephp\cakephp\src\Error\ErrorHandler.php on line 178
```
This is due to a missing return in validation($error) in Crud\Error\ExceptionRenderer. If you add 
```
return $this->controller->response; 
```
everything works fine and validation errors are getting displayed.

sean-nicholas:
Fixes #228

### DIFF
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -20,7 +20,7 @@ class ExceptionRenderer extends \Cake\Error\ExceptionRenderer {
  * Renders validation errors and sends a 412 error code
  *
  * @param \Exception $error Exception instance
- * @return void
+ * @return \Cake\Network\Response
  */
 	public function validation($error) {
 		$url = $this->controller->request->here();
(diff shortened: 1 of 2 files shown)


ADmad:
Closing in favor of #242. Thank you for reporting the issue and helping to fix it.

PR #237, opened 2015-01-04
update deprecated method

antograssiot:
``String`` Class has been replace by ``Text``
``validatePresence()`` has been renamed ``requirePresence()``


jippi:
<3

ADmad:
I am surprised there are no fails in spite of the core changes to validation. Not sure if that's good or bad :stuck_out_tongue:

antograssiot:
also @Admad the tests where failing due to the String Clas call despite of the class_alias. Do you think it is a phpunit issue only ?

ADmad:
That should work for tests too. It probably means the file which adds the class alias is not being loaded during tests.

PR #238, opened 2015-01-04
Cake3 - Options

ADmad:


src/Action/AddAction.php
```diff
@@ -50,10 +51,8 @@ class AddAction extends BaseAction
         'viewVar' => null,
         'relatedModels' => true,
         'entityKey' => 'entity',
-        'saveOptions' => [
-            'validate' => true,
-            'atomic' => true
-        ],
+        'entityOptions' => [],
+        'saveOptions' => [],
```
lorenzo: It is actually a good practice to pass the same options to patchEntity and save. So having a single array should be fine

jippi: would it make sense to have `options` and then one for each specialized ?

lorenzo: Well, all the work I've done in the marshaller and save have been to encourage people to use a single array for options to avoid duplication and to also drive some automation out of it.

ADmad: @lorenzo Oh that's nice. If the same options can be used for both then I too am in favor of not adding a separate key for it. 

@jippi The list of config keys it reasonably long already. If adding new keys can be avoided we should.

jippi: Ok, I'm fine with just one key :)

ADmad:
Removed entityOptions

lorenzo:
Good job!

PR #239, opened 2015-01-08
3.0 - Update add action to prevent showing validation errors on GET request.

dakota:
Due to the change on the validation layer in CakePHP, validation errors are shown on a empty GET request.


lorenzo:
@dakota what happens if I post an empty array?

dakota:
@lorenzo That'll give validation messages atm. It depends on how we'd want to treat an empty POST?

lorenzo:
It should give validation errors, so that's perfect

PR #242, opened 2015-01-27
Fix API exception rendering

jadb:
As it was, the `ExceptionRenderer` was not being registered by the listener
and the `Cake\Network\Response` was not properly returned on validation
exception rendering.

### DIFF
--- a/src/Listener/ApiListener.php
+++ b/src/Listener/ApiListener.php
@@ -152,8 +152,8 @@ class ApiListener extends BaseListener
      */
     public function registerExceptionHandler()
     {
-        $config = ['exceptionRenderer' => 'Crud\Error\ExceptionRenderer'];
-        (new ErrorHandler($config))->register();
+        $exceptionRenderer = 'Crud\Error\ExceptionRenderer';
+        (new ErrorHandler(compact('exceptionRenderer') + (array)Configure::read('Error')))->register();
     }
 
     /**
--- a/src/Listener/ApiListener.php
+++ b/src/Listener/ApiListener.php
@@ -2,7 +2,7 @@
 namespace Crud\Listener;
 
 use Cake\Core\Configure;
-use Cake\Error\Handler;
+use Cake\Error\ErrorHandler;
 use Cake\Event\Event;
 use Cake\Network\Exception\BadRequestException;
 use Cake\Network\Request;
--- a/src/Listener/ApiListener.php
+++ b/src/Listener/ApiListener.php
@@ -2,6 +2,7 @@
 namespace Crud\Listener;
 
 use Cake\Core\Configure;
+use Cake\Error\Handler;
 use Cake\Event\Event;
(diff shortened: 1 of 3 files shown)


src/Listener/ApiListener.php
```diff
@@ -2,6 +2,7 @@
 namespace Crud\Listener;
 
 use Cake\Core\Configure;
+use Cake\Error\Handler;
```
ADmad: Handler => ErrorHandler

src/Listener/ApiListener.php
```diff
@@ -151,7 +152,8 @@ protected function _checkRequestMethods()
      */
     public function registerExceptionHandler()
     {
-        Configure::write('Error.exceptionRenderer', 'Crud\Error\ExceptionRenderer');
+        $config = ['exceptionRenderer' => 'Crud\Error\ExceptionRenderer'];
```
ADmad: The problem with this is any other non-default settings are lost. There's an open PR/dicussion related to this cakephp/app#183

ADmad:
Related #229

src/Listener/ApiListener.php
jadb: derp - that's what happens when I try quickly re-writing something...

src/Listener/ApiListener.php
jadb: Then it shouldn't be called `registerExceptionHandler()`, and the registration should be left to the user. It just mixes up to see it there, no mention in the docs and then have to dig to figure it out.

In any case, it wasn't working, even when manually registered in bootstrap for validation errors because of the error with `ExceptionRenderer::validation()`.

jadb: How about something like this:

```php
(new ErrorHandler($config + (array)Configure::read('Error')))->register();
```

People would be required to change the `Configure::consume()` in app to `Configure::read()` until a better solution is found there? As it is, it's broken or requires more work than just this change to one's bootstrap if using the `ApiListener`. What do you think?

ADmad: That sounds good for now.

tests/TestCase/Listener/ApiListenerTest.php
```diff
@@ -1027,40 +1027,40 @@ public function testSetupDetectorsIntigration()
...
-            ->disableOriginalConstructor()
-            ->getMock();
-
-        $listener->registerExceptionHandler();
-
-        $expected = 'Crud\Error\ExceptionRenderer';
-        $result = Configure::read('Error.exceptionRenderer');
-        $this->assertEquals($expected, $result);
-    }
+    // public function testRegisterExceptionHandlerWithApi()
```
jippi: just remove dead code or fix it ;)

jadb: Didn't want to just drop the test and say it was useless ;)

PR #243, opened 2015-01-27
Hide exception details

jadb:
Exception info should only be shown in debug mode.

### DIFF
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -133,16 +133,13 @@ class ExceptionRenderer extends \Cake\Error\ExceptionRenderer
             }
         }
 
-        if (!empty($viewVars['error'])) {
+        if (!empty($viewVars['error']) && Configure::read('debug')) {
             $data['exception'] = [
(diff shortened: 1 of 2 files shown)


ADmad:
Test failing

jippi:
:-1: I prefer having some exception data leaked in non-debug mode, basically everything that does not reveal anything useful about the app, but allow the end-user to easier debug the issue - e.g. custom security exceptions, preconditions etc

jadb:
@jippi what data exactly? 

https://gist.github.com/jadb/8aa5defd2081316e6f38

According to the gist above, when the trace is not included, the message and error code are already part of the data (outside of exception), so the only data leaked is the  exception renderer class. Have I missed something?

jadb:
I'll correct myself, it shows the Validation exception class. But still, to an end-user, it doesn't give more information than the message/code it includes in the parent data already:

```
Crud\\Error\\Exception\\ValidationException
```
 and 

```
A validation error occurred
```

What do you think?

jippi:
Seems legit :)

PR #245, opened 2015-02-02
Infinity json_encode()

kimegede:
Saw this happening with `ApiTransformationListener` when one of my fields got casted to number.

Test showing casting a string there will result being casted to infinity (`INF`) will fail when json_encoded().
[See `JSON_ERROR_INF_OR_NAN`](http://php.net/manual/en/function.json-last-error.php)

### DIFF
--- a/Controller/Component/CrudComponent.php
+++ b/Controller/Component/CrudComponent.php
@@ -96,7 +96,7 @@ class CrudComponent extends Component {
  * By default no actions are enabled.
  *
  * `listeners` List of internal-name => ${plugin}.${class} listeners
- * that will be bound automatically in Crud. By default the related model event
+ * that will be bound automatically in Crud. By default the related models' events
  * are bound. Events will always assume to be in the Controller/Event folder.
  *
  * `eventLogging` boolean to determine whether the class should log triggered events.
@@ -272,7 +272,7 @@ class CrudComponent extends Component {
 /**
  * Map the view file to use for a controller action.
  *
- * To map multiple action views in one go pass an array as first argument and no second argument.
+ * To map multiple action views in one go pass an array as the first argument with no second argument.
  *
  * @param string|array $action
  * @param string $view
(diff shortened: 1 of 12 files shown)


jadb:
@kimegede sorry, I don't use the cake2 version myself - maybe someone else will be able to chime in, add a fix to this PR and merge everything.

kimegede:
I did make a fix, not super happy about it.... but seems to work.
So hoped to get some feedback - https://github.com/FriendsOfCake/crud/pull/245/files#diff-e8c4daffe7ff3824fad7e1c8283e28abR305

jippi:
I'm not sure what the actual problem is @kimegede :)

kimegede:
The string "0113870020e12235" will be seen as a numeric, and then get converted.
But the converting will result in INF(inity) and INF can't be json_encode'd.

```php
is_numeric('0113870020e12235') // true
"0113870020e12235" + 0 // INF
var_dump(json_encode(['foo' => INF]); // false
echo json_last_error(); // 7 - "One or more NAN or INF values in the value to be encoded" http://php.net/json_last_error
```

So in my case, that resulted in a white screen. Nothing in PHP log files or anything.

Controller/Crud/Listener/ApiTransformationListener.php
```diff
@@ -302,7 +302,11 @@ protected function _castNumbers($variable) {
 		if (!is_numeric($variable)) {
 			return $variable;
 		}
-		return $variable + 0;
+
+		$hintTyped = $variable + 0;
```
AD7six: would be better named "$number"

AD7six:
:+1: in principle.

Relevant test fails on 5.2 and 5.3 though.

kimegede:
@AD7six as I understand Travis-ci, it's saying there was stuff there was broken before my changes.
And looks like it's analyzing each commit of mine too? (First time with Travis-ci, not sure if I misunderstand the result)

AD7six:
@kimegede [look at the 5.2 run](https://travis-ci.org/FriendsOfCake/crud/jobs/49288392#L251), and the [5.3 run](https://travis-ci.org/FriendsOfCake/crud/jobs/49288393#L366). That test didn't exist before your changes, since it's in this PR =).

jippi:
There is some failed unit tests?

PR #249, opened 2015-03-24
Merge component's message config with action's message config.

ADmad:
This change for e.g. allows you to set "params" options just once instead of having to do it for every message type for every action.

One can do `$this->Crud->config('messages.params.class' => 'alert-message')` and have the class used for all messages.

### DIFF
--- a/src/Action/BaseAction.php
+++ b/src/Action/BaseAction.php
@@ -125,16 +125,27 @@ abstract class BaseAction extends Object
 
         $crud = $this->_crud();
 
-        $config = $this->config('messages.' . $type);
-        if (empty($config)) {
-            $config = $crud->config('messages.' . $type);
-            if (empty($config)) {
-                throw new \Exception(sprintf('Invalid message type "%s"', $type));
-            }
+        $crudConfig = $crud->config('messages.' . $type);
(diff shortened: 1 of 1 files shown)


jadb:
I don't mind this but here's how I was doing it using events:

```php
    protected function _flashMessagesSetup()
    {
        $this->Crud->on('setFlash', function (Event $event) {
            $event->subject->params['class'] = 'alert-message';
        });
    }
```

ADmad:
That's 3 lines too many for me to write in my app  :stuck_out_tongue:

ADmad:
Seriously though I am reconsidering whether it's worth it to merge the options this way.

lorenzo:
I prefer the even based approach. But having a test would be nice to speed the merge of this feature

ADmad:
Using code below achieves what I want and let's BootstrapUI's FlashHelper handle the css class. So I am going to be lazy and close this for now rather than adding test cases for the change :stuck_out_tongue: 

```php
        $this->Crud->on('setFlash', function (Event $event) {
            unset($event->subject->params['class']);
            $event->subject->element = ltrim($event->subject->type);
        });
```

Issue #250, opened 2015-04-07
Scrutinizer dependency check fails on cake3 crud


bravo-kernel:
Hi guys, I hope this is not caused by a configuration error on my side but my scrutinizer fails on (just) crud. All other dependencies work as expected (including other C3 plugins).

```
  Notice: Undefined index: cake\event\event in vendor/scrutinizer/php-analyzer/src/Scrutinizer/PhpAnalyzer/Model/Repository/EntityTypeProvider.php line 131  
```

The details can be found [here](https://scrutinizer-ci.com/g/alt3/cakephp-app-plusplus/inspections/c988a912-939b-4dd4-a0e3-6b4acf2ca2cc). Just click on the red cross.

jadb:
Where is `cake\event\event` used without property capitalization? Looks like its related to something like that.

bravo-kernel:
I will check my plugin for any lower casing in my plugin tomorrow. Would be akward to say the least 😏

bravo-kernel:
Just checked and there are only two references to Cake\Event\Event in my (almost empty) plugin and they are using the **correct casing**, just like found in the CRUD plugin:

```php
use Cake\Event\Event;

    /**
     * beforeFilter
     *
     * @param Cake\Event\Event $event Event instance
     * @return void
     */
    public function beforeFilter(Event $event)
    {
        pr("AppPlusPlus component beforeFilter()");
    }
```

Could it be related to the CRUD plugin and not my code? No biggie if this isn't solved, I can just disable Scrutinizer for the time being.

bravo-kernel:
Spotted one difference though, CRUD uses a leading \, testing right now.

```php
* @param \Cake\Event\Event $event Event instance
```

jippi:
feel free to PR a fix for it :)

bravo-kernel:
Sorry guys, I have triple-checked my sources and even disabled everything remotely related to this but the outcome is still the same and because the Scrutinizer information is so limited I honestly have no idea what to investigate next.

bravo-kernel:
I have stripped my app to the bare minimum but the issue persists. Since this plugin is pretty rock-solid the issue could very well be caused by some configuration f*** up on my behalf so... feel free to close if you guys think this is one of those cases.

dakota:
For the record, I am getting the exact same issue.

dakota:
Asked Scrutinizer support, and they said:

[...]

PR #253, opened 2015-04-14
Fix docblock param types

dakota:
I think that this may correct #250. Will need to merge and see what happens.

### DIFF
--- a/src/Controller/Component/CrudComponent.php
+++ b/src/Controller/Component/CrudComponent.php
@@ -193,7 +193,7 @@ class CrudComponent extends Component
     /**
      * Called after the Controller::beforeFilter() and before the controller action.
      *
-     * @param Cake\Event\Event $event Event instance
+     * @param \Cake\Event\Event $event Event instance
      * @return void
      */
     public function startup(Event $event)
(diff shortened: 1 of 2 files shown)


ADmad:
It doesn't :)  I really doubt scrutinizer depends on API docblocks tags.

ADmad:
Oops I deleted the bugfix branch but here's the link to failed scrutinizer build https://scrutinizer-ci.com/g/FriendsOfCake/crud/inspections/0edf8859-87d9-493c-90d0-be7ece2b4fa0

dakota:
I'm specifically refering to the failure with projects that use crud (As reported in #250) - https://scrutinizer-ci.com/g/alt3/cakephp-app-plusplus/inspections/c988a912-939b-4dd4-a0e3-6b4acf2ca2cc (Click on the `friendsofcake/crud` link)

The exact error reported there is:

```php
  [ErrorException]                                                                                                                                                   
  Notice: Undefined index: cake\event\event in vendor/scrutinizer/php-analyzer/src/Scrutinizer/PhpAnalyzer/Model/Repository/EntityTypeProvider.php line 131 
```

dakota:
On second thought, the two errors are probably related and caused by the same thing.

PR #261, opened 2015-04-29
Initial pass at a BulkAction base handler

josegonzalez:
This will allow us to implement `_bulk` handlers for actions that occur from `index`-like pages. Any subclass that implements a `_bulk` method will be given a $query object that can be used to reference a set of database objects. The classes can then do whatever they like - such as publish all posts, or delete records.

This sort of method is more useful for admin modules than for apis, but is a common task in application development

### DIFF
--- a/docs/actions.rst
+++ b/docs/actions.rst
@@ -142,4 +142,8 @@ More On Actions
 	actions/add
 	actions/edit
 	actions/delete
+  actions/bulk
+  actions/bulk-delete
+  actions/bulk-set-value
+  actions/bulk-toggle
 	actions/custom
--- a/docs/actions.rst
+++ b/docs/actions.rst
@@ -142,8 +142,8 @@ More On Actions
 	actions/add
 	actions/edit
 	actions/delete
-  actions/bulk
-  actions/bulk-delete
-  actions/bulk-set-value
-  actions/bulk-toggle
+	actions/bulk
+	actions/bulk-delete
+	actions/bulk-set-value
+	actions/bulk-toggle
 	actions/custom
--- a/src/Action/Bulk/ToggleAction.php
+++ b/src/Action/Bulk/ToggleAction.php
@@ -58,7 +58,7 @@ class ToggleAction extends BaseAction
     protected function _bulk(Query $query = null)
     {
         $field = $this->config('field');
-        $expression = [new QueryExpression(sprintf('%s = NOT %s', $field))];
+        $expression = [new QueryExpression(sprintf('%1$s= NOT %1$s', $field))];
         $query->update()->set($expression);
         $statement = $query->execute();
         $statement->closeCursor();
--- a/src/Action/Bulk/ToggleAction.php
+++ b/src/Action/Bulk/ToggleAction.php
@@ -3,7 +3,7 @@ namespace Crud\Action\Bulk;
 
 use Cake\Controller\Controller;
 use Cake\Database\Expression\QueryExpression;
-use Cake\Network\Exception\NotImplementedException;
+use Crud\Error\Exception\ActionNotConfiguredException;
 use Cake\ORM\Query;
 
 /**
@@ -45,7 +45,7 @@ class ToggleAction extends BaseAction
     {
         $field = $this->config('field');
         if (empty($field)) {
-            throw new NotImplementedException('No field value specified');
+            throw new ActionNotConfiguredException('No field value specified');
         }
 
         return parent::_handle();
--- a/src/Action/Bulk/ToggleAction.php
+++ b/src/Action/Bulk/ToggleAction.php
@@ -3,8 +3,8 @@ namespace Crud\Action\Bulk;
 
 use Cake\Controller\Controller;
 use Cake\Database\Expression\QueryExpression;
-use Crud\Error\Exception\ActionNotConfiguredException;
 use Cake\ORM\Query;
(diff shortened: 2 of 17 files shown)


josegonzalez:
Comments welcome! Obviously I have no tests, but thats because I'd like to see the implementation be fleshed out more before we continue.

src/Action/BulkAction.php
```diff
@@ -0,0 +1,90 @@
...
+     *
+     * @return void
+     */
+    protected function _handle()
+    {
+        if (!method_exists($this, '_bulk')) {
+            throw new NotImplementedException(sprintf(
+                'Action %s does not implement a _bulk handler',
+                get_class($this)
+            ));
```
lorenzo: It would be better to just make `abstract function _bulk` instead of dynamically checking on each request.

src/Action/BulkAction.php
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        $ids = $this->_controller()->request->data('id');
```
lorenzo: should check that $ids is actually an array

src/Action/BulkAction.php
josegonzalez: Heh I added it as abstract later, guess I can remove this check.

src/Action/BulkAction.php
```diff
@@ -0,0 +1,90 @@
...
+
+        $alias = $this->_table()->alias();
+        $primaryKey = $this->_table()->primaryKey();
+        if (is_array($primaryKey)) {
+            $name = $this->config('action');
+            throw new ActionNotConfiguredException(sprintf('Action "%s" does not have a proper findConfig', $name));
+        }
+
+        $config['conditions'] = [
+            "{$alias}.{$primaryKey}" => $subject->ids,
```
lorenzo: Should be `"{$alias}.{$primaryKey} IN" => $subject->ids`

lorenzo:
Will there be a default implementation of this action?

josegonzalez:
No, there won't be. Thats why it's abstract?

ADmad:
Can provide a "delete all".

lorenzo:
@josegonzalez There should be at least one, imo. Otherwise this will go unnoticed,  may I suggest bulk delete?

josegonzalez:
I can do that.

@lorenzo re: your comment on ensuring the `$ids` is an array, what should I do if it's not?

ADmad:
Another could be toggling of boolean fields.

src/Action/BulkAction.php
```diff
@@ -0,0 +1,83 @@
...
+        if (!empty($config)) {
+            return $config;
+        }
+
+        $alias = $this->_table()->alias();
+        $primaryKey = $this->_table()->primaryKey();
+        if (is_array($primaryKey)) {
+            $name = $this->config('action');
+            throw new ActionNotConfiguredException(sprintf('Action "%s" does not have a proper findConfig', $name));
+        }
```
lorenzo: There is an easy way of supporting composite keys if we wanted to:

```php
$config['conditions'] = function ($exp) use ($primaryKey, $ids) {
    return $exp->in($primaryKey, $ids);
```

lorenzo:
@josegonzalez throw an BadRequest exception

josegonzalez:
Okay I think these all work, though I'd appreciate it if someone else wrote tests :D . I can definitely write docs as well.

I've tested these all in a dashboard with crud-view, seems to work fine :)

src/Action/Bulk/ToggleAction.php
```diff
@@ -0,0 +1,68 @@
...
+        return parent::_handle();
+    }
+
+    /**
+     * Handle a bulk toggle
+     *
+     * @param \Cake\ORM\Query $query The query to act upon
+     * @return bool
+     */
+    protected function _bulk(Query $query = null)
```
AD7six: I think bulk-toggle is not a good an idea.

In principle it's fine - but in practice it means when the action runs you won't have any idea what it'll actually do since it depends on the db state at the time. For example, accidentally double-submit and it "doesn't work at all". Better to use explicit/idempotent actions.

That said, it's providing a tool (rope) for devs to use. Any other thoughts along the above lines?

AD7six:
My one comment aside - :+1:

src/Action/Bulk/ToggleAction.php
ADmad: As you said we are just providing a tool here. Let's leave it to the devs to decide it's appropriate for their needs. If that  column is being shown on the index page table it does give a visual indication of the updated values.

josegonzalez: Yeah, what @ADmad said. If someone wants to be *more* clear about their application implementation, there is also `SetValueAction`.

josegonzalez:
Okay I made a first pass at writing tests. I guess.

josegonzalez:
Can this be merged meow?

docs/actions.rst
```diff
@@ -142,4 +142,8 @@ More On Actions
 	actions/add
 	actions/edit
 	actions/delete
+  actions/bulk
+  actions/bulk-delete
+  actions/bulk-set-value
+  actions/bulk-toggle
```
lorenzo: Bad indentation

AD7six: I can fix that - the previous lines are tab-indented, is that right?

PR #264, opened 2015-05-06
Properly call _trigger() with a Subject object

josegonzalez:
We previously were sending it an array.

### DIFF
--- a/src/Controller/Component/CrudComponent.php
+++ b/src/Controller/Component/CrudComponent.php
@@ -486,7 +486,7 @@ class CrudComponent extends Component
      * object.
      *
      * @param string $eventName Event name
-     * @param \Cake\Event\Subject $subject Event data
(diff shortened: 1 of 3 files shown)


josegonzalez:
Fixes an issue produced by #263.

ADmad:
Tests please.

josegonzalez:
Sorry no, I'm not going to write tests for this. #263 was the thing that broke functionality - sans tests! - and was merged without complaints. I'll write a test for this when a test is written for the other stuff.

ADmad:
Going by that logic we should stop bothering with tests henceforth. Everyone is going to make a mistake sometime, so we will just assign the blame and move on.

josegonzalez:
What would the test be? That php throws an error if we pass in an array instead of a subject?

jippi:
Yes

PR #270, opened 2015-05-26
Update README.md

davidyell:
Changed the version notice so that it marks this branch as for 3.x


davidyell:
Urgh, my reading skills fail. Although I bet I'm not the first to assume that `3.0 => 3.x` instead of `4.0 => 3.x`

PR #273, opened 2015-06-03
Use associationClass name instead of association

GuidoHendriks:
Related models used association name as the key for associations. This value is lowercase model name without underscores, like it's set in associations. I changed it so it uses the association's name, this way it's a proper camel cased model name.

### DIFF
--- a/src/Listener/RelatedModelsListener.php
+++ b/src/Listener/RelatedModelsListener.php
@@ -117,7 +117,7 @@ class RelatedModelsListener extends BaseListener
                 continue;
             }
 
-            $return[$association] = $associationClass;
+            $return[$associationClass->name()] = $associationClass;
         }
 
         return $return;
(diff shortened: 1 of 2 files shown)


ADmad:
Can you please add a test case for this.

ADmad:
What benefit does this change have?  From what I see the key isn't used anywhere.

GuidoHendriks:
It is used when calling `$this->Crud->on('relatedModel', ...)`. I discussed this issue with @lorenzo on IRC.

ADmad:
Ah right, I misread the code. It's used [here](https://github.com/GuidoHendriks/crud/blob/fix-association-name/src/Listener/RelatedModelsListener.php#L53).

GuidoHendriks:
If you don't define your own models, it won't work with the example on this page: http://crud.readthedocs.org/en/latest/listeners/related-models.html

jippi:
can you please submit a test case for this issue?

GuidoHendriks:
I could try that, but it won't be until next week at best..

jippi:
Sure, we'll wait :)

GuidoHendriks:
I added the tests.

ADmad:
:+1:

Issue #283, opened 2015-06-24
Running the tests with phpunit 4.7.x


davidyell:
I was updating the tests and I noticed that if you define the `modelClass` in the controller Crud will automatically try and access the table inflected from the controller, rather than the table defined in the Table class.

```php
// src/Controller/ExamplesController.php
class ExamplesController extends AppController {
    public function initialize()
    {
        $this->modelClass = 'Kittens';
    }

// src/Model/Table/KittensTable.php
class Kittens extends Table {
    public function initialize(array $config)
    {
        $this->table('kittens');
    }
}
```

Crud, by default, will try and find the `examples` table in the database, rather than reading the `modelClass`.

The [example of the setup in the tests is here](https://github.com/FriendsOfCake/crud/blob/master/tests/TestCase/Controller/Component/CrudComponentTest.php#L60),  as [this table uses the `posts` table](https://github.com/FriendsOfCake/crud/blob/master/tests/App/Model/Table/CrudExamplesTable.php#L32) but the test fails because the `crud_examples` table doesn't exist.

Be sure to run the tests with `phpunit/phpunit` `~4.7`

ADmad:
> The example of the setup in the tests is here, as this table uses the posts table but the test fails because the crud_examples table doesn't exist.

Running the testsuite I don't get any failures.

davidyell:
@ADmad  You don't even get a deprecated warning for the fact that `assertTag()` has been removed from phpunit?

jippi:
travis is happy - all is good?

davidyell:
Travis is happy with `PHPUnit 3.7.33 by Sebastian Bergmann.`

ADmad:
Until someone finds the time to update the use of deprecated/removed methods like `assertTag()` it's gonna keep using 3.7.

davidyell:
Yep, that was the main thing which tripped me up, and the guy replacing the assertions hasn't updated his code.

PR #284, opened 2015-06-24
Replace deprecated eventManager methods

davidyell:
I've updated the component to use the new `on` and `off` methods as the old ones have been deprecated. As such I've done my best to update the main component test so that it passes.

Although there is what I think is a bug, which I've detailed in #283 

The Action integration tests are also broken due to `assertTag()` having been deprecated. There is a replacement in the works, but the methods have not been ported yet. More information, see https://github.com/phpunit/phpunit-dom-assertions/issues/3

### DIFF
--- a/src/Controller/Component/CrudComponent.php
+++ b/src/Controller/Component/CrudComponent.php
@@ -411,7 +411,7 @@ class CrudComponent extends Component
                 $event = $this->_config['eventPrefix'] . '.' . $event;
             }
 
-            $this->_eventManager->attach($callback, $event, $options);
+            $this->_eventManager->on($event, $options, $callback);
         }
     }
 
@@ -468,7 +468,7 @@ class CrudComponent extends Component
         }
 
         if (isset($this->_listenerInstances[$name])) {
-            $this->_eventManager->detach($this->_listenerInstances[$name]);
+            $this->_eventManager->off($this->_listenerInstances[$name]);
             unset($this->_listenerInstances[$name]);
         }
 
@@ -657,7 +657,7 @@ class CrudComponent extends Component
             unset($config['className']);
             $this->_listenerInstances[$name]->config($config);
 
-            $this->_eventManager->attach($this->_listenerInstances[$name]);
+            $this->_eventManager->on($this->_listenerInstances[$name]);
 
             if (is_callable([$this->_listenerInstances[$name], 'setup'])) {
                 $this->_listenerInstances[$name]->setup();


composer.json
```diff
@@ -43,7 +43,7 @@
         "cakephp/cakephp": "~3.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "3.7.33",
+        "phpunit/phpunit": "~4.7",
```
ADmad: Just change it to "*", that's what core uses.

davidyell:
Due to the tests being incompatible with the latest phpunit, let me know if you'd like me to retarget my PR to a branch for development, rather than master, so hopefully some brainier people can pitch in to update some of the other tests. :+1:

ADmad:
Don't skip tests. Revert back the phpunit version change.

ADmad:
Keep your changes limited to only updating the eventmanager calls in this PR.

davidyell:
Sorry, I don't understand.

If the phpunit version is reverted to `3.7.33` then the PR can be closed, as the test updates are only needed for `4.7.x` of phpunit.

Using phpunit 4.7 the `Action` tests will fail on a deprecated warning for `assertTag`, so by marking them as skipped would hopefully allow the other tests to be worked on without the suite failing whilst waiting for the new `phpunit/phpunit-dom-assertion` to be released, or the tests could be reworked.

Changes to the eventManager bind methods will still cause errors in the `Action` tests, so let me know if you'd like me to omit these from this PR. I had trouble updating the test to correctly mock the new `\Cake\Controller\Component\FlashComponent`

ADmad:
> Changes to the eventManager bind methods will still cause errors in the Action tests,

They won't if you don't change the phpunit version.

davidyell:
Let me know if this is okay. Also is it worth squashing some of these down?

ADmad:
Are the changes to `CrudComponentTest` directly related to event manger method changes?

There are phpcs errors https://travis-ci.org/FriendsOfCake/crud/jobs/68165119#L195

davidyell:
No, they are not, they were made to get the test to pass in phpunit `4.7.x`. I was running the test suite to ensure that the changes I'd made still passed the tests before I submitted my PR. I think the phpcs errors are because I haven't properly reset that test file.

I can fix up those phpcs errors no problem.

ADmad:
> No, they are not, they were made to get the test to pass in phpunit 4.7.x

As I said earlier only change whats required to update the eventmanger calls. Revert back all other unrelated changes.

davidyell:
I've squashed them all down into a single commit, so I'll close this and re-open a cleaner PR.

davidyell:
Seems github is too clever for me, it's updated its self.

PR #285, opened 2015-06-24
Replaced the deprecated EventManager methods

davidyell:
Updated event manager methods in the component.

### DIFF
--- a/src/Action/BaseAction.php
+++ b/src/Action/BaseAction.php
@@ -52,13 +52,13 @@ abstract class BaseAction extends Object
 
         if (method_exists($this, $method)) {
             $this->_responding = true;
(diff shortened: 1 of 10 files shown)


ADmad:
`BaseAction` and various test classes also use `attach()` instead of `on()` which should be updated.

davidyell:
Haha, yeah, that's what got me embroiled in all that stuff with the tests to start with!

ADmad:
Just concentrate on task at hand and don't muck around with other things :)

ADmad:
Good job @davidyell :)

davidyell:
:clap: 
Glad to help, got there in the end, thanks for the guidance.

PR #290, opened 2015-06-28
Remove verson in readme's install command.

ADmad:
This way the latest stable release will be installed.


ADmad:
Hmm that installs dev-master if the apps composer.json has `"minimum-stability": "dev"`

PR #295, opened 2015-07-06
Fixed issue caused by validation of query string

lorenzo:
When displaying an add form and sending some of the fields in the query string,
it is better to not pre-validate the full form as it will obviously contain errors.


jadb:
:+1:

PR #301, opened 2015-07-19
Pass model factory to be used

jadb:


src/Action/BaseAction.php
```diff
@@ -26,6 +26,15 @@
     protected $_responding = false;
 
     /**
+     * Default configuration
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'modelFactory' => 'Table',
```
ADmad: Every existing action declares `$_defaultConfig` so this is going to be overwritten. That's why all builds are failing.

jadb: Hmm, right - worked in my case, but hadn't run the tests. Do you agree with the change? Should I go over all the actions?

ADmad: It will still cause problem for custom actions people have. You can instead check if the key is set in ProxyTrait and use if available.

jippi: :+1: for @ADmad 's suggestion 

Also remember to update the docs with the new option :)

I know @lorenzo also is working on getting crud / crud-view to play nice with ElasticSearch :)

jadb:
@ADmad done. @jippi I am so bad at docs... please feel free to add to this PR :) I'll submit my 'hacks' (not sure it's the way they want them) for ES - have it working here with Crud (at least for the tests i have written thus far).

lorenzo:
Great!

PR #303, opened 2015-07-20
Fix a bug where Api Pagination does not work for plugins.

AndersTrier:
When ApiPaginationListener is used from a plugin, $controller->modelClass is set to 'pluginname.model', which then fails on "array_key_exists($modelClass, $request->paging)".


jippi:
Hi, can you please provide a test for this case? :)

PR #304, opened 2015-07-21
Resolves #287

### LINKED ISSUE #287: Auto-loading related Tables
**Scenario**
I've made my admin index view and baked some views. It so happens that this table Examples, has a `status_id` field which relates to the Statuses table. This data isn't loaded.

**What I did**
In my AppController I made a method which I called from my `beforeFilter`.

```php
protected function addRelatedTables()
{
    if (!is_array($this->paginate['contain']) || !isset($this->paginate['contain'])) {
        $this->paginate['contain'] = [];
    }

    $this->Crud->on('beforePaginate', function (Event $event) {
        foreach($this->{$this->modelClass}->associations() as $table => $association) {
            $this->paginate['contain'] = array_merge(
                $this->paginate['contain'],
                [$this->{$this->modelClass}->association($table)->name()]
            );
        }
    });
}
```

**Suggestions**
Is it worth making this into a listener in a similar vein to the RelatedModelsListener?

**Concerns**
With so many ways to configure the Paginator component, my code example above relies on the shortcut pagination array, and I don't know if it would be merged accurately with someone using the paginator component directly, such as `$this->Paginator->paginate($Table, $optionsArray)`.

Also I'd be interested in thoughts on how configure such a thing in a more application wide way, to prevent controller specific configuration, which, to me, defeats the point of the Crud plugin.

davidyell:
Added a `beforePaginate` event hook to the RelatedModelsListener so that paginated data will contain related table classes.

### DIFF
--- a/src/Listener/RelatedModelsListener.php
+++ b/src/Listener/RelatedModelsListener.php
@@ -27,6 +27,19 @@ class RelatedModelsListener extends BaseListener
     }
 
     /**
+     * Automatically parse and contain related table classes
+     *
+     * @param Event $event
+     * @return void
+     */
+    public function beforePaginate(Event $event)
+    {
+        if (!$event->subject()->query->contain()) {
+            $event->subject()->query->contain($this->relatedModels());
+        }
+    }
+
+    /**
      * Find and publish all related models to the view
      * for an action
      *


davidyell:
Quite busy at work, but I'll see if I can get a test working for this.

davidyell:
The configuration I used to get it working was as follows.

```php
        $this->loadComponent('Crud.Crud', [
            'actions' => [
                'index' => [
                    'className' => 'Crud.Index',
                    'relatedModels' => ['Users']
                ],
                'view' => [
                    'className' => 'Crud.View',
                    'relatedModels' => ['Users', 'Tags']
                ]
            ]
        ]);
```

ADmad:
Tests?

davidyell:
I did head you off at the pass this time @ADmad :wink:

jippi:
looks good once its tested ! :)

davidyell:
Sorry, but I've spent 5.5 hours trying to write a test for this :cyclone:, and I've been unsuccessful. So I'm going to stop here. 

Here are the two approaches that I've tried, along with the issues that I experienced when trying to get the test to pass. Hopefully it's helpful in some way.

**First effort**
```php
    public function testContainingRelatedTables()
    {
        $this->markTestIncomplete('This test relies on having a return in the event.');

        $listener = $this
            ->getMockBuilder('\Crud\Listener\RelatedModelsListener')
            ->disableOriginalConstructor()
            ->setMethods(['relatedModels'])
            ->getMock();

        $listener
            ->expects($this->once())
            ->method('relatedModels')
            ->will($this->returnValue([
                'Users' => [
                    'Posts'
                ],
                'Tags'
            ]));

        $db = $this->getMockBuilder('\Cake\Database\Connection')
            ->disableOriginalConstructor()
            ->getMock();

        $query = new Query($db, null);

        $subject = new Subject(['query' => $query]);
        $event = new Event('beforePaginate', $subject);

        $result = $listener->beforePaginate($event);
        $contain = $result->query->contain();

        $this->assertNotNull($contain);
        $this->assertArrayHasKey('Users', $contain);
        $this->assertArrayHasKey('Tags', $contain);
        $this->assertArrayHasKey('Posts', $contain['Users']);
    }
```
This test will pass, but it relies on making changes to the event, in order that the subject can be returned. If you do not return the subject, there is no way, using this approach, to inspect the Query object.

The change is minor, and I haven't tested it in an application to see if it has any side effects.
```php
    public function beforePaginate(Event $event)
    {
        if (!$event->subject()->query->contain()) {
            $event->subject()->query->contain($this->relatedModels());
            return $event->subject();
        }
    }
```

**Second effort**
In the second attempt I tried to approach it from more of an integration test standpoint so that I could try and interrogate the returned data. This was because after talking to @jippi, he said that the `beforePaginate` event doesn't return anything, not event an `$event->result`. So the only way to properly test would be to actually trigger the event from a `paginate()` call.
```php
    public function testRelatedTableLoading()
    {
        $controller = new \Crud\Test\App\Controller\BlogsController();
        $controller->Blogs->hasMany('Tags');

        $controller->Crud->config('actions.index', ['className' => 'Crud.Index', 'relatedModels' => ['Tags']]);

        $controller->eventManager()->on('Crud.beforePaginate', new \Crud\Listener\RelatedModelsListener($controller));

        $this->loadFixtures('Blogs', 'Tags');
        $q = $controller->paginate($controller->Blogs);

//        $reflect = new \ReflectionClass($q);
//        $property = $reflect->getProperty('_query');
//        $property->setAccessible(true);
//
//        var_dump($property->getValue($reflect));

//        var_dump($q->toArray());

    }
```
The issue I had trying to get this working was two fold. Firstly, `paginate()` returns a `Cake\ORM\ResultSet` which does have the Query in `_query` but unfortunately I was unable to reflect the property to check the contain.

So as you can see my second stab was to try and join the Blogs fixture to a Tags fixture and look for the associated data after the paginate call. However, calling `paginate()` directly sidesteps the Crud event handler, and thus the `Crud.beforePaginate` event is never fired, so the listener never has a chance to change the `contain()`.

Short of writing a full-on integration test, which I'm not familiar with, I'm not sure how else to proceed.

Thanks to @hackzilla, @jippi and @lorenzo for the help with this! Karma++ :gift_heart:

[I've pushed the code up to my fork over here](https://github.com/davidyell/crud/blob/287-tests/tests/TestCase/Listener/RelatedModelsListenerTest.php#L243). So if someone wants to adapt the tests from that branch, please feel free.

davidyell:
It'd be great if someone could take a stab at this.

davidyell:
Any chance of someone picking up this test @jippi ?

davidyell:
Thanks @admad

PR #309, opened 2015-07-24
Added section about disabling already loaded actions in a controller

Spriz:
I havn't compiled this, so I don't know if it looks correct - otherwise I'm sure somebody could point me where to go :smile:

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -132,9 +132,11 @@ Example of disable a loaded action:
   }
 
 ..code-block:: phpinline
-  class AppController extends \Cake\Controller\Controller {
+  class PostsController extends AppController {
 
     public function beforeFilter(\Cake\Event\Event $event) {
+      parent::beforeFilter($event);
+
       $this->Crud->disable(['Edit', 'Delete']);
     }
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -114,6 +114,32 @@ An example of on the fly enabling an Crud action with configuration
 
   }
 
+Disabling loaded actions
+----------------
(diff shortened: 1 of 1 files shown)


docs/configuration.rst
```diff
@@ -114,6 +114,32 @@ An example of on the fly enabling an Crud action with configuration
 
   }
 
+Disabling loaded actions
+----------------
```
jippi: the `-` need to cover the entire length of the title

docs/configuration.rst
```diff
@@ -114,6 +114,32 @@ An example of on the fly enabling an Crud action with configuration
...
+    public $components = [
+      'Crud.Crud' => [
+        'actions' => ['Crud.Index', 'Crud.View', 'Crud.Delete', 'Crud.Edit']
+      ]
+    ];
+
+  }
+
+..code-block:: phpinline
+  class AppController extends \Cake\Controller\Controller {
```
jippi: this one should not be AppController - but an instance extending the AppController

docs/configuration.rst
```diff
@@ -114,6 +114,32 @@ An example of on the fly enabling an Crud action with configuration
...
+
+    public function beforeFilter(\Cake\Event\Event $event) {
+      $this->Crud->disable(['Edit', 'Delete']);
```
jippi: should also call parent :)

Spriz: Fixed

docs/configuration.rst
Spriz: Fixed.

docs/configuration.rst
Spriz: Fixed.

PR #311, opened 2015-08-09
Throw exception when 'relatedModels' association not found

deizel:
Borrows same exception/message from core as a more helpful alternative to:

    Fatal error: Call to a member function name() on null


jadb:
:+1:

PR #312, opened 2015-08-11
related-models docs

steefaan:
Before the docs improvement it's not clear that the listener needs to be loaded. It sounds like that `relatedModels` will work without any additional configuration. Especially because the following explanation:

```
By default all related model lists for main CRUD table instance will be fetched, but only for add, edit and corresponding admin actions.
```


jadb:
Thanks :+1:

PR #315, opened 2015-08-15
init search listener

steefaan:
FOC/Search meets FOC/Crud.

An implementation of an additonal listener to play nicely with FOC/Search.

Test cases not done yet, comments are welcome.

EDIT:
Don't merge yet, the test case is incomplete.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "3.7.33",
-        "friendsofcake/cakephp-test-utilities": "dev-master"
+        "friendsofcake/cakephp-test-utilities": "dev-master",
+        "friendsofcake/search": "dev-master"
     },
     "autoload": {
         "psr-4": {
@@ -57,7 +58,7 @@
         }
     },
     "suggest": {
-        "friendsofcake/search": "Searchable Crud.Index or whenever you use beforePaginate.",
+        "friendsofcake/search": "Provides search capabilities for the Crud plugin.",
         "friendsofcake/crud-view": "Automated admin backend based on your Crud configuration."
     },
     "support":{
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,8 @@
         }
     },
     "suggest": {
-        "friendsofcake/crud-view": "Automated admin backend based on your Crud configuration."  
+        "friendsofcake/search": "Searchable Crud.Index or whenever you use beforePaginate.",
+        "friendsofcake/crud-view": "Automated admin backend based on your Crud configuration."
     },
     "support":{
         "source":"https://github.com/FriendsOfCake/crud",
--- a/docs/listeners/search.rst
+++ b/docs/listeners/search.rst
@@ -1,7 +1,7 @@
 Search listener
 =================
 
-Searchable Crud.Index or whenever you use ``beforePaginate``.
+Provides search capabilities for the Crud plugin.
 
 Setup
 -----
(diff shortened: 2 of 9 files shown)


src/Listener/SearchListener.php
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Returns a list of all events that will fire in the controller during its lifecycle.
+     * You can override this function to add your own listener callbacks
+     *
+     * @return array
+     */
+    public function implementedEvents()
+    {
+        return [
+            'Crud.beforePaginate' => ['callable' => 'beforePaginate']
```
steefaan: Not sure about the priority some inputs?

composer.json
```diff
@@ -57,7 +57,8 @@
         }
     },
     "suggest": {
-        "friendsofcake/crud-view": "Automated admin backend based on your Crud configuration."  
+        "friendsofcake/search": "Searchable Crud.Index or whenever you use beforePaginate.",
```
steefaan: Maybe not the best explanation. Maybe someone has a better one?

josegonzalez: `Provides search capabilities for the Crud plugin`

docs/listeners/search.rst
```diff
@@ -0,0 +1,43 @@
...
+
+Attach it on the fly in your controller beforeFilter, this is recommended if
+you want to attach it only to specific controllers and actions:
+
+.. code-block:: php
+
+  <?php
+  class SamplesController extends AppController {
+
+    public function beforeFilter() {
```
ADmad: Event argument missing.

tests/TestCase/Listener/SearchListenerTest.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    /**
+     * Test beforePaginate
+     *
+     * @return void
+     */
+    public function testBeforePaginate()
+    {
+
```
steefaan: Unsure here:

1. Is it necessary to test because it's more or less `FOC/Search` land.
2. Testable? Because it's only a composer suggestion, it means the only thing I could test is the `RumetimeException`.

src/Listener/SearchListener.php
lorenzo: No need to mess with priorities

src/Listener/SearchListener.php
```diff
@@ -0,0 +1,41 @@
...
+        ];
+    }
+
+    /**
+     * Inject search conditions to the qeury object.
+     *
+     * @param \Cake\Event\Event $event
+     * @return void
+     */
+    public function beforePaginate(Event $event)
```
lorenzo: I would rename this method and make it a bit more reusable, since it can also be handy for `beforeLookup`

tests/TestCase/Listener/SearchListenerTest.php
lorenzo: you will need to add the search plugin in `require-dev`

src/Listener/SearchListener.php
steefaan: Rename yes, but what you mean with make it more reusable?

lorenzo: So that both `beofrePaginate` and `beforeLookup` can be handled by the same method; and having configuration in the listener so that you can disable one of the events if needed

tests/TestCase/Listener/SearchListenerTest.php
```diff
@@ -0,0 +1,85 @@
...
+            ->will($this->returnValue([
+                'search' => [
+                    'name' => '1st post'
+                ]
+            ]));
+
+        $blogs = \Cake\ORM\TableRegistry::get('Blogs');
+        $subject->query = $blogs->find();
+
+
```
steefaan: Does anyone know how to continue here? I need to trigger the event and chack the result afterwardsm then the PR is done.

steefaan:
Thanks to @korotovsky.

@lorenzo You're fine with the test case? The PR should be done now.

lorenzo:
Perfect... thanks!

PR #316, opened 2015-08-21
Fix for "Fatal Error: Class 'String' not found" in CakePHP 2.7.x

Asgraf:
Following line causes "Fatal Error: Class 'String' not found" in CakePHP 2.7.x
https://github.com/FriendsOfCake/crud/blob/3.0/Controller/Crud/CrudAction.php#L239
This patch fixes the problem

### DIFF
--- a/Controller/Crud/CrudAction.php
+++ b/Controller/Crud/CrudAction.php
@@ -3,6 +3,7 @@
 App::uses('CrudBaseObject', 'Crud.Controller/Crud');
 App::uses('Hash', 'Utility');
 App::uses('Validation', 'Utility');
+App::uses('String', 'Utility');
 
 /**
  * Base Crud class


Asgraf:
https://github.com/FriendsOfCake/crud/blob/3.0/Controller/Crud/CrudAction.php#L239
This line causes Fatal Error in CakePHP 2.7.x

josegonzalez:
We should probably use `CakeText` - or whatever it became - instead.

Asgraf:
@josegonzalez -  i thought of same thing but then realised that CakePHP 2.7 is not intended for new projects.
It is intended as migration step from CakePHP 2.x to 3.x

Use of CakeText in crud/3.0 branch would brake compatibility with all existing CakePHP 2.x projects that hasn't yet been upgraded to 2.7.x

PR #317, opened 2015-08-26
Isit forgoten to be updated?

hkosmidi:
I use the latest crud plugin but all $event->subject->item doesn't exist. But $event->subject->entity does. Is it a mispell in docs? Or forgoten to be updated?


lorenzo:
Thanks, it was a mistake

PR #321, opened 2015-09-17
optional flash messages

steefaan:
In general I agree that flash messages are not necessary for api requests. Let me explain with the following use case why an optional flash message may be useful.

I want to use `deleteAll` in a normal cake template context. Regarding issues with `<form>` in `<form>` I decided to execute `deleteAll` with an AJAX request. Therefore I loaded the API listener to get a well-formed response for my AJAX request. Now I want to refresh my page in jQuery's `success` callback and want to show a flash message to the user after the reload is done.

What do you think?

EDIT:
As usual, test cases follow after you're fine with the changes itself.

### DIFF
--- a/src/Listener/ApiListener.php
+++ b/src/Listener/ApiListener.php
@@ -32,6 +32,7 @@ class ApiListener extends BaseListener
             'json' => 'Json',
             'xml' => 'Xml'
         ],
+        'setFlash' => false,
         'detectors' => [
             'json' => ['ext' => 'json', 'accepts' => 'application/json'],
             'xml' => ['ext' => 'xml', 'accepts' => 'text/xml']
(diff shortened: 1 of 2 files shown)


src/Listener/ApiListener.php
```diff
@@ -390,7 +391,9 @@ public function viewClass($type, $class = null)
      */
     public function setFlash(Event $event)
     {
-        $event->stopPropagation();
+        if (!$this->config('setFlash')) {
```
ADmad: This is stop the event even when `setFlash` is undefined, which will be the case when the `ApiListener` is not attached and existing apps. Instead explicitly check if value is `false`.

ADmad:
You can go ahead and update tests :)

steefaan:
@ADmad Done.

steefaan:
Something missing or can be merged?

ADmad:
@steefaan Looks good to me. Was waiting for a +1 from one more FOC member but I will merge later today if there isn't.

lorenzo:
+1

PR #322, opened 2015-09-21
deprecated method viewClassMap

steefaan:
refs http://book.cakephp.org/3.0/en/appendices/3-1-migration-guide.html#requesthandlercomponent

Has anyone an idea why the other test cases fail? It should be working like before without to change the test cases.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         }
     ],
     "require":{
-        "cakephp/cakephp": "~3.0"
+        "cakephp/cakephp": "~3.1"
     },
     "require-dev": {
         "phpunit/phpunit": "3.7.33",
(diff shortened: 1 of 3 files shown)


lorenzo:
Can you also change composer.json to require cakephp 3.1 as minimum?

ADmad:
Builds are still failing

steefaan:
Yes @lorenzo will take a look on it later.

steefaan:
@ADmad Not sure if this should be merged but probably not a problem because this fixes at least the deprecation error. Not sure why the build fails, it seems that the ApiListener itself works fine.

ADmad:
@steefaan The failing tests are due to changes in ExceptionRender in 3.1, unrelated to the fix you made.

PR #323, opened 2015-09-21
Simplify ExceptionRenderer::_outputMessage()

ADmad:
Set trace info in response only if not already added by
core's ExceptionRenderer::render()

### DIFF
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -4,9 +4,9 @@ namespace Crud\Error;
 use Cake\Core\Configure;
 use Cake\Core\Exception\MissingPluginException;
 use Cake\Datasource\ConnectionManager;
+use Cake\Error\Debugger;
 use Cake\Event\Event;
 use Cake\View\Exception\MissingViewException;
-use Cake\Error\Debugger;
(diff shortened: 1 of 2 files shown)


src/Error/ExceptionRenderer.php
```diff
@@ -119,8 +97,14 @@ protected function _getErrorData()
                 'class' => get_class($viewVars['error']),
                 'code' => $viewVars['error']->getCode(),
                 'message' => $viewVars['error']->getMessage(),
-                'trace' => preg_split('@\n@', $viewVars['error']->getTraceAsString()),
             ];
+
+            if (!isset($data['trace'])) {
```
jippi: the trace now doesn't exist under the exception key?

ADmad: Since 3.1's exception renderer already sets [trace](https://github.com/cakephp/cakephp/blob/master/src/Error/ExceptionRenderer.php#L169) it's present in `$data['trace']` unless you have a custom method for an exception. So I set it under `$data` itself to maintain consistency.

jippi: alright

PR #324, opened 2015-09-22
Use custom finder defined in association for related models

ankr:


jippi:
wouldn't it make sense to be able to override the 'list' param ?

lorenzo:
@jippi That would be a different feature... a good one, actually... but unrelated to what he is trying to fix

PR #325, opened 2015-09-23
related models conditions

steefaan:
If you define a relation like:

```
$this->belongsTo('Categories', [
    'conditions' => [
        'type' => 'products'
    ]
]);
```

in your table class, the related models listener does not consider these conditions.


davidyell:
Good find @steefaan ! :+1:

lorenzo:
Fixed yesterday already: https://github.com/FriendsOfCake/crud/pull/324

PR #326, opened 2015-09-23
delete second find()

steefaan:
refs https://github.com/FriendsOfCake/crud/pull/324

### DIFF
--- a/src/Listener/RelatedModelsListener.php
+++ b/src/Listener/RelatedModelsListener.php
@@ -51,7 +51,7 @@ class RelatedModelsListener extends BaseListener
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


lorenzo:
A bug in cake needs to be fixed before this PR is merged

steefaan:
@lorenzo Already works for me with a single `find()` in Cake 3.1

jadb:
I agree with @steefaan  - been working for me for a while (since `3.1.x-dev`).

lorenzo:
There is still a bug, I will try to fix it tomorrow in cakephp

ADmad:
This is no longer an issue on current master.

PR #329, opened 2015-10-01
Use "treeList" finder if model has TreeBehavior loaded.

ADmad:


jippi:
icanhaztest?

ADmad:
@jippi Hows that look?

src/Listener/RelatedModelsListener.php
```diff
@@ -51,7 +52,8 @@ public function publishRelatedModels($action = null)
                 continue;
             }
 
-            $query = $association->find()->find('list');
+            $finder = $this->_finder($association);
```
deizel: Undefined method - was this intended to be protected, or public like in the tests?

ADmad: @deizel It should be public. I have pushed a fix.

deizel: :+1:

PR #332, opened 2015-10-20
Removed default class passed to flash

Wouter0100:
Removed adding `$type` as class.

This because the `class` could also be an array. And, in some cases we don't want to add `$type` to it the array, can't we do this better manually in the config?

This change makes it also compatible with the `bootstrap-ui` plugin (I was using both plugins, which, didn't work properly without making these changes to `crud`)

### DIFF
--- a/src/Action/BaseAction.php
+++ b/src/Action/BaseAction.php
@@ -146,7 +146,6 @@ abstract class BaseAction extends Object
 
         $config = Hash::merge([
             'element' => 'default',
-            'params' => ['class' => 'message'],
             'key' => 'flash',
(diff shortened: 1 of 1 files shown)


Wouter0100:
Updated, now I didn't just remove the defaults, but when `class` isn't false, it will use the defaults. On this way we're able to say to the plugin that it must not do anything with the `class`.

ADmad:
Can't you just use the `Crud.setFlash` event to modify the flash settings as per your needs?

Wouter0100:
Yeaah, that was possible too.. Hehe..

PR #335, opened 2015-10-29
Contain related models in beforePaginate()

ADmad:
Refs #287 #304

### DIFF
--- a/src/Listener/RelatedModelsListener.php
+++ b/src/Listener/RelatedModelsListener.php
@@ -17,6 +17,40 @@ class RelatedModelsListener extends BaseListener
 {
 
     /**
+     * Returns a list of events this listener is interested in.
+     *
+     * @return array
+     */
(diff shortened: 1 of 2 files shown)


src/Listener/RelatedModelsListener.php
```diff
@@ -17,6 +17,40 @@ class RelatedModelsListener extends BaseListener
...
+     * @return void
+     */
+    public function beforePaginate(Event $event)
+    {
+        $contained = $event->subject()->query->contain();
+        if (!empty($contained)) {
+            return;
+        }
+
+        $models = $this->models();
```
lorenzo: what is this method?

lorenzo: Nevermind, I thought this was an action class.

davidyell:
I know this is posthumous but it's not actually related to #304 at all.

Issue #341, opened 2015-11-10
ApiQueryLog somehow asks for datasources that doesn't exist


Spriz:
In my cake 3.0.15 project we have different datasources: `default`, `test` and `alpha`.

If I enable Crud like this:
```
$this->loadComponent('Crud.Crud', [
            'actions' => [
                'Crud.Index',
                'Crud.Add',
                'Crud.Edit',
                'Crud.View',
                'Crud.Delete'
            ],
            'listeners' => [
                'Crud.Api',
                'Crud.ApiPagination',
                'Crud.ApiQueryLog'
            ]
        ]);
```

and in my Integration test uses following:

```
$this->get('someUrl/:id);
$this->assertResponseOk();
```

I get the following error:

```
PHPUnit 5.0.8 by Sebastian Bergmann and contributors.


Starting test 'Api\Test\TestCase\Controller\WallPostsControllerTest::testIndex'.
F                                                                   1 / 1 (100%)

Time: 473 ms, Memory: 17.00Mb

There was 1 failure:

1) Api\Test\TestCase\Controller\WallPostsControllerTest::testIndex
exception 'Cake\Datasource\Exception\MissingDatasourceConfigException' with message 'The datasource configuration "test_alpha" was not found.' in /var/www/dev/apacta-2/vendor/cakephp/cakephp/src/Datasource/ConnectionManager.php:187
Stack trace:
#0 /var/www/dev/apacta-2/vendor/friendsofcake/crud/src/Listener/ApiQueryLogListener.php(109): Cake\Datasource\ConnectionManager::get('alpha')
#1 /var/www/dev/apacta-2/vendor/friendsofcake/crud/src/Listener/ApiQueryLogListener.php(51): Crud\Listener\ApiQueryLogListener->_getSource('alpha')
#2 /var/www/dev/apacta-2/vendor/cakephp/cakephp/src/Event/EventManager.php(385): Crud\Listener\ApiQueryLogListener->setupLogging(Object(Cake\Event\Event))
#3 /var/www/dev/apacta-2/vendor/cakephp/cakephp/src/Event/EventManager.php(355): Cake\Event\EventManager->_callListener(Array, Object(Cake\Event\Event))
#4 /var/www/dev/apacta-2/vendor/friendsofcake/crud/src/Controller/Component/CrudComponent.php(505): Cake\Event\EventManager->dispatch(Object(Cake\Event\Event))
[...]

PR #342, opened 2015-11-11
Catch MissingDatasourceConfigException in ApiQueryLogListenern

Spriz:
refs #341

### DIFF
--- a/src/Listener/ApiQueryLogListener.php
+++ b/src/Listener/ApiQueryLogListener.php
@@ -106,6 +106,10 @@ class ApiQueryLogListener extends BaseListener
      */
     protected function _getSource($source)
     {
-        return ConnectionManager::get($source);
+        try {
+            return ConnectionManager::get($source);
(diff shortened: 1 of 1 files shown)


src/Listener/ApiQueryLogListener.php
```diff
@@ -106,6 +106,10 @@ protected function _getSources()
      */
     protected function _getSource($source)
     {
-        return ConnectionManager::get($source);
+        try {
+            return ConnectionManager::get($source);
+        } catch (\Cake\Datasource\Exception\MissingDatasourceConfigException $e) {
```
lorenzo: This is not the best place to ignore... The best place is where this method is called from

Spriz: rather want these to lines in try block? https://github.com/Apacta/crud/blob/edf3a9e1a2d3e1523429e72f75d7fb7b75aff259/src/Listener/ApiQueryLogListener.php#L51-L52

lorenzo:
@Spriz correct!

Issue #338, opened 2015-10-31
Limit DeleteAction GET


steefaan:
Other than expected it's possible to delete content with a normal GET request. I think we should limit this to POST/DELETE as suggested also for the normal cake delete action. What do you think?

jippi:
You can it by either checking the action + request type in the controller, or only allow the router to access /delete using HTTP DELETE requests :)

lorenzo:
I think this is not a bad idea to have by default

adayth:
I was here to post the same issue. In Crud 3.0 was limited to DELETE / POST requests. I think the same behavior should be OK.

adayth:
Meanwhile you could use a code in AppController like this to limit access to delete actions mapped with crud:
```php
public function beforeFilter(\Cake\Event\Event $event) 
{
        parent::beforeFilter($event);

        // Limit HTTP valid queries access to delete actions mapped with CRUD
        if ($this->request->action === 'delete'
            && $this->Crud->isActionMapped('delete')
            && !$this->request->is(['post', 'delete'])) {
            throw new MethodNotAllowedException();
        }
    }
```

steefaan:
@adayth This is quite simpler:

```
        if (in_array($this->request->action, ['delete'])) {
            $this->request->allowMethod(['post', 'delete']);
        }
```

adayth:
Thanks @steefaan !

ADmad:
@steefaan Why use `in_array` for a static array with just one value? :)

bravo-kernel:
:tractor:

steefaan:
@ADmad Sorry was a copy paste from something like:

```
        if (in_array($this->request->action, ['index', 'lookup'])) {
            $this->Crud->addListener('Crud.Search');
            $this->loadComponent('Search.Prg');
        }
```

But yes you're right:

```
if ($this->request->action === 'delete') {
    $this->request->allowMethod(['post', 'delete']);
}
```

is enough here.

ADmad:
Done in #346

PR #344, opened 2015-11-11
limit delete action methods

steefaan:
refs https://github.com/FriendsOfCake/crud/issues/338

### DIFF
--- a/src/Action/DeleteAction.php
+++ b/src/Action/DeleteAction.php
@@ -49,12 +49,12 @@ class DeleteAction extends BaseAction
     ];
 
     /**
-     * HTTP DELETE handler
+     * HTTP POST handler
      *
      * @param string $id Record id
      * @return void
      */
-    protected function _handle($id = null)
+    protected function _post($id = null)
     {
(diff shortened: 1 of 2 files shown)


lorenzo:
Thanks!

jippi:
I really dislike this - it prevents GET for /delete/ - its not even possible to opt-in to it - so its a huge breaking change - I'm going to revert it until we can come up with a way to introduce it that won't break existing apps and allow for people to opt in to GET, even though its not recommended

josegonzalez:
:+1: what @jippi said. That's something I thought might be weird. Maybe just a configuration value and a check in `_handle` for it?

lorenzo:
@jippi We can just release a new tag and let people convert their apps. It makes absolutely no sense to have a delete over GET, IMO

josegonzalez:
@lorenzo I would do it for a fallback with a View to let people confirm the delete first.

lorenzo:
@josegonzalez with the previous implementation it will delete before showing the view

PR #345, opened 2015-11-11
Revert "limit delete action methods"

jippi:
Reverts FriendsOfCake/crud#344


lorenzo:
wtf?

lorenzo:
So anyone using crud in a public app and having links to delete will get all their data removed by a google crawl... by default

lorenzo:
@jippi The cake 2 version did not work like this https://github.com/FriendsOfCake/crud/blob/develop/Controller/Crud/Action/DeleteCrudAction.php

PR #347, opened 2015-11-14
Catch exception and ignore it in ApiQueryLogListener

Spriz:
Refs #341

### DIFF
--- a/src/Listener/ApiQueryLogListener.php
+++ b/src/Listener/ApiQueryLogListener.php
@@ -48,8 +48,12 @@ class ApiQueryLogListener extends BaseListener
     public function setupLogging(Event $event)
     {
         foreach ($this->_getSources() as $connectionName) {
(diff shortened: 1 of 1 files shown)


lorenzo:
Thanks

ADmad:
I am curious how do we end up with non-existent datasources when `_getSources()` uses `ConnectionManager::configured()` to get the list.

lorenzo:
The aliasing done by the fixture manager when running tests

ADmad:
@lorenzo Aren't the params to `ConnectionManager::alias()` swapped [here](https://github.com/cakephp/cakephp/blob/master/src/TestSuite/Fixture/FixtureManager.php#L141)?  As per API the first param is supposed to be actual name and second param the alias.

lorenzo:
@ADmad No... they order is misleading to me too. But the order is correct.

PR #350, opened 2015-11-20
adding the missing return

pedrofs:
Hey,

This missing return statement makes the Response to be null so no redirect is made.

I couldn't figure out how the tests are passing for `delete` method.

Best,
Pedro.

### DIFF
--- a/src/Action/DeleteAction.php
+++ b/src/Action/DeleteAction.php
@@ -83,7 +83,7 @@ class DeleteAction extends BaseAction
      */
     protected function _delete($id = null)
     {
-        $this->_post($id);
+        return $this->_post($id);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ADmad:
Could you add a test to avoid regression in future.

pedrofs:
@ADmad I had to remove the `delete.ctp` template view to get the tests failing right.

lorenzo:
Thanks

lorenzo:
@pedrofs If you have time to add a test so that we prove the current behavior is the right one, we would appreciate it

pedrofs:
@lorenzo after removing the `delete.ctp` it ensures the right behavior, isn't it?

lorenzo:
@pedrofs only by chance (or omission) hehe

PR #351, opened 2015-11-24
Corrects incorrect afterSave() subject key name

johanmeiring:
The `afterSave()` event documentation indicates that `item` is one of the keys of the event `subject()`.  This is incorrect.  The correct name of the key is `entity`.  This PR makes the necessary correction to the documentation.


lorenzo:
Thanks

PR #353, opened 2015-11-26
Update installation.rst

simkimsia:
bump to 4.2

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -23,7 +23,7 @@ Using the inline require for composer:
 
 .. code-block:: sh
 
-	composer require friendsofcake/crud:~4.0
+	composer require friendsofcake/crud:~4.2
 
 Or add this to your composer.json configuration:
 
@@ -31,7 +31,7 @@ Or add this to your composer.json configuration:
 
 	{
 		"require" : {
-			"FriendsOfCake/crud": "~4.0"
+			"FriendsOfCake/crud": "~4.2"
 		}
 	}
 


ADmad:
Better to just remove the version as it will automatically fetch the latest release. Only the install command for CakPHP 2.x needs to explicitly mention version.

docs/installation.rst
```diff
@@ -31,7 +31,7 @@ Or add this to your composer.json configuration:
 
 	{
 		"require" : {
-			"FriendsOfCake/crud": "~4.0"
+			"FriendsOfCake/crud": "~4.2"
```
ADmad: Can change `"~4.2" to `"@stable"

lorenzo: I would consider `@stable` dangerous

simkimsia: @lorenzo, it would be good if you can write a 1 liner explaining why it's dangerous.

Also, I was wondering if there is a way to update the version number in 1 place and auto show up in all the README.md and .rst files

lorenzo: Because if we go to version 5, that will be the stable version. But your project actually depends on version 4

docs/installation.rst
```diff
@@ -23,7 +23,7 @@ Using the inline require for composer:
 
 .. code-block:: sh
 
-	composer require friendsofcake/crud:~4.0
+	composer require friendsofcake/crud:~4.2
```
dakota: This can actually just be `composer require friendsofcake/crud` composer will automatically infer the correct latest version.

docs/installation.rst
dakota: I would actually remove this, directly editing the composer.json file is discouraged (Lest we waken the [DOHMSKRAKEN](https://twitter.com/dohmsraken)) - [source](http://blog.doh.ms/2014/10/13/installing-composer-packages/)

PR #355, opened 2015-11-26
Splits POST and PUT actions

bravo-kernel:
Splits POST and PUT actions as they respond differently, also removes:

- note about HTTP header (with link to /view) being sent after succesful POST
- mention of 301 code for POST (since it does not seem to be used)


lorenzo:
thanks

PR #358, opened 2015-11-27
Make API DELETE description consistent with other actions

bravo-kernel:


simkimsia:
Why isn't this merged yet?

PR #359, opened 2015-11-27
Bake command now uses template rather than view

lornajane:


josegonzalez:
FWIW I *think* you can just add `[ci skip]` to docs changes (I know CircleCI supports it, not sure if TravisCI does as well).

PR #361, opened 2015-11-28
A better way to bake controller for Crud

simkimsia:
This works well when you type

`bin/cake bake controller users --theme Crud`

which will properly create the add action and a beforeFilter

### DIFF
--- a/src/Template/Bake/Element/add.ctp
+++ b/src/Template/Bake/Element/add.ctp
@@ -1,0 +1,25 @@
+<%
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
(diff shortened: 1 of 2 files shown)


simkimsia:
reminder to self:

- ask how to write tests for the baked controller
- write docs on how to use the `bin/cake bake` for this

src/Template/Bake/Element/add.ctp
```diff
@@ -0,0 +1,25 @@
...
+%>
+
+    /**
+     * Add method
+     *
+     * @return void Redirects on successful add, renders view otherwise.
+     */
+    public function add()
+    {
+        return (new \Crud\Action\AddAction($this))->handle();
```
raul338: Again it should be
```php
```

simkimsia: But I emulate from https://github.com/FriendsOfCake/crud-demo-app/blob/master/App/Controller/PostsController.php#L46

lorenzo: That demo is incorrect, though. It should be as @raul338 pointed out

lorenzo: I have another demo here https://github.com/lorenzo/cakephp3-bookmarkr

josegonzalez: I fixed the demo :)

simkimsia:
I will cancel this PR and redo another one.

PR #365, opened 2015-11-29
An improved better way to bake the controller for add action

simkimsia:
This PR overrides the #361 PR

### DIFF
--- a/src/Template/Bake/Controller/controller.ctp
+++ b/src/Template/Bake/Controller/controller.ctp
@@ -7,6 +7,7 @@ $defaultModel = $name;
 namespace <%= $namespace %>\Controller<%= $prefix %>;
 
 use <%= $namespace %>\Controller\AppController;
+<% if (!empty($actions)) : %>use Cake\Event\Event;<% endif; %>
 
 /**
  * <%= $name %> Controller
@@ -22,6 +23,22 @@ foreach ($components as $component):
 class <%= $name %>Controller extends AppController
 {
 <%
+echo $this->Bake->arrayProperty('helpers', $helpers, ['indent' => false]);
 echo $this->Bake->arrayProperty('components', $components, ['indent' => false]);
+%>
+<% if (!empty($actions)) : %>
+    public function beforeFilter(Event $event) {
+        parent::beforeFilter($event);
+
+        // you can remove all these mapActions if your AppController has mapped the actions via configuration
+        // read http://crud.readthedocs.org/en/latest/configuration.html#actions for details
+        // e.g. of a on-the-fly enable action
+        // $this->Crud->mapAction('add', 'Crud.Add');
+    }
+<% endif; %>
+<%
+foreach($actions as $action) {
+    echo $this->element('Controller/' . $action);
+}
 %>
 }
--- a/src/Template/Bake/Element/add.ctp
+++ b/src/Template/Bake/Element/add.ctp
@@ -1,0 +1,52 @@
+<%
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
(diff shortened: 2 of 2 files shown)


simkimsia:
Once this is accepted, I have another PR that bakes the edit action as well.

Currently, I take the cue of how to bake the add action based on the cakephp/bake template for add action

```
    /**
     * Add method
     *
     * @return void Redirects on successful add, renders view otherwise.
     */
    public function add()
    {
        $<%= $singularName %> = $this-><%= $currentModelName %>->newEntity();
        if ($this->request->is('post')) {
            $<%= $singularName %> = $this-><%= $currentModelName %>->patchEntity($<%= $singularName %>, $this->request->data);
            if ($this-><%= $currentModelName; %>->save($<%= $singularName %>)) {
                $this->Flash->success(__('The <%= strtolower($singularHumanName) %> has been saved.'));
                return $this->redirect(['action' => 'index']);
            } else {
                $this->Flash->error(__('The <%= strtolower($singularHumanName) %> could not be saved. Please, try again.'));
            }
        }
<%
        $associations = array_merge(
            $this->Bake->aliasExtractor($modelObj, 'BelongsTo'),
            $this->Bake->aliasExtractor($modelObj, 'BelongsToMany')
        );
        foreach ($associations as $assoc):
            $association = $modelObj->association($assoc);
            $otherName = $association->target()->alias();
            $otherPlural = $this->_variableName($otherName);
%>
        $<%= $otherPlural %> = $this-><%= $currentModelName %>-><%= $otherName %>->find('list', ['limit' => 200]);
<%
            $compact[] = "'$otherPlural'";
        endforeach;
%>
        $this->set(compact(<%= join(', ', $compact) %>));
        $this->set('_serialize', ['<%=$singularName%>']);
    }

```

All the set and serialize for the primary model is settled by the AddAction already.

All the set for the related models is settled by the publishRelatedModels function **provided** the relatedModels listener is turned on.

What is missing is setting the find('list') on the relatedmodels to limit 200 hence I added a change for on `relatedModel` event

raul338:
The PR code looks fine (didn't test it)

As for the Add, you don't need the newEntity() because it is handled by Crud. And for the _serialize you can use the Crud API Listener

simkimsia:
@raul338 Ah I see. Thanks for the tip!

src/Template/Bake/Controller/controller.ctp
```diff
@@ -22,6 +23,22 @@ foreach ($components as $component):
...
+
+        // you can remove all these mapActions if your AppController has mapped the actions via configuration
+        // read http://crud.readthedocs.org/en/latest/configuration.html#actions for details
+        // e.g. of a on-the-fly enable action
+        // $this->Crud->mapAction('add', 'Crud.Add');
+    }
+<% endif; %>
+<%
+foreach($actions as $action) {
+    echo $this->element('Controller/' . $action);
```
josegonzalez: I think if the template doesn't exist, this will throw an exception, no?

simkimsia: If this template does not exist, I believe that the bake shell will default to use the cakephp/bake add template.

I have tested by deliberately not having a Bake folder inside FriendsOfCake/crud/src/Template and then still run `bin/cake bake controller users_in_apps --theme Crud`

I got a Error: "Crud" is not a valid value for --theme. Please use one of "Bake, Migrations"

I have tested by deliberately not having a `add.ctp` inside FriendsOfCake/crud/src/Template/Bake/Element/ and then still run `bin/cake bake controller users_in_apps --theme Crud`

the add action baked was that of the original one we expect from cakephp/bake.

Does that help to answer your question, @josegonzalez ?

simkimsia: Anything else I need to answer to make this PR work? @josegonzalez

josegonzalez: Well if you get an error, then this bake theme doesn't work, right?

jippi:
Closing it as the PR haven't gotten any love for quite some time

PR #373, opened 2015-12-18
Combine AddAction and EditAction logic into BaseCreateAction

josegonzalez:
This change provides a single interface into creating/updating
records. The logic from both classes was highly similar, with only
the retrieval of the initial GET subject and the entity being posted
being different.

One possibility this change introduces is the ability to have a
`CreateOrUpdateAction`, wherein the entity will be created regardless
of it's initial existence. Users can simply subclass the
`BaseCreateAction` class and define the two methods necessary to
handle the cases they wish to handle.

### DIFF
--- a/src/Action/BaseCreateAction.php
+++ b/src/Action/BaseCreateAction.php
@@ -116,7 +116,7 @@ abstract class BaseCreateAction extends BaseAction
      *
      * @param \Crud\Event\Subject $subject Event subject
      * @param mixed $id Record id
-     * @return void
+     * @return \Crud\Event\Subject
      */
     abstract protected function _getSubject(Subject $subject, $id = null);
 
@@ -125,7 +125,7 @@ abstract class BaseCreateAction extends BaseAction
      *
      * @param \Crud\Event\Subject $subject Event subject
      * @param mixed $id Record id
-     * @return void
+     * @return \Cake\ORM\Entity
      */
     abstract protected function _postEntity(Subject $subject, $id = null);
 }
--- a/src/Action/BaseCreateAction.php
+++ b/src/Action/BaseCreateAction.php
@@ -1,0 +1,131 @@
+<?php
+namespace Crud\Action;
+
+use Crud\Event\Subject;
+use Crud\Traits\FindMethodTrait;
+use Crud\Traits\RedirectTrait;
+use Crud\Traits\SaveMethodTrait;
+use Crud\Traits\SerializeTrait;
+use Crud\Traits\ViewTrait;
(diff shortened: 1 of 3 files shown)


josegonzalez:
Thoughts on this? I didn't want to just merge it without any comments, but it seems like it's good to go and doesn't break any tests so...

lorenzo:
I just skimmed the code... Will this break my code if I had extended before the Add or Edit action?

josegonzalez:
I don't *think* so, though its possible you used undocumented/untested functionality. Do you have an example of a class where this might be problematic?

josegonzalez:
@lorenzo *bump* would love to get this in.

jippi:
whats the benefit of this change @josegonzalez ?

jhli:
Another way would be to split up into GET add, GET edit, PUT create, PATCH update. See rails3 ressource routes. See http://www.informit.com/articles/article.aspx?p=1671632&seqNum=11

josegonzalez:
> One possibility this change introduces is the ability to have a
CreateOrUpdateAction, wherein the entity will be created regardless
of it's initial existence. Users can simply subclass the
BaseCreateAction class and define the two methods necessary to
handle the cases they wish to handle.

lorenzo:
Looks good to me

src/Action/BaseCreateAction.php
```diff
@@ -0,0 +1,131 @@
...
+use Crud\Traits\ViewTrait;
+use Crud\Traits\ViewVarTrait;
+
+/**
+ * Base Create/Update Crud class
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ */
+abstract class BaseCreateAction extends BaseAction
```
jippi: i dislike the `BaseCreateAction` name, especially when an edit class suddenly extends a Create class - doesn't make sense to me - currently I'm loss for a better name, but BaseCreateAction wont fly imo

thinkingmedia:
This hides the functionality of the AddAction and forces developers to refer to the base class in order to inspect it's functionality. It turns AddAction into a micro class that doesn't perform the work of adding.

The BaseCreateAction becomes a utility class that forces the user to inherit. Now when ever BaseCreateAction changes those inherits classes could break.

I like having action classes that perform a specific thing and being able to go to that code as a reference.

jippi:
I'm generally not keen on this change either, mostly because things quickly gets messy

why not have a dedicated action for CreateOrUpdate ? the boilerplate for that is quite slim anyway - or make a feature flag for `update`  to allow it to create on NotFound for a record

PR #384, opened 2016-01-26
Fix incorrect subject key name in docs

johanmeiring:
The documentation incorrectly indicated that a key named `item` was available as part of the event subject for the `beforeRedirect` and `setFlash` events.  The correct name for the key is `entity`.  This PR corrects the documentation.


ADmad:
Thanks.

PR #385, opened 2016-02-01
Adding Crud::list action

ankr:
Like `index` action but without pagination. So you don't have to do `limit => 9999` to get a list of all your categories (or other table).

### DIFF
--- a/src/Traits/FindMethodTrait.php
+++ b/src/Traits/FindMethodTrait.php
@@ -56,6 +56,35 @@ trait FindMethodTrait
     }
 
     /**
+     * Find all records
+     *
+     * @param \Crud\Event\Subject $subject Event subject
+     * @return \Cake\ORM\Entity
+     */
+    protected function _findAll(Subject $subject)
+    {
+        $repository = $this->_table();
+
+        $query = $repository->find($this->findMethod());
+
+        $subject->set([
+            'repository' => $repository,
+            'query' => $query
+        ]);
+
+        $this->_trigger('beforeFind', $subject);
+        $entities = $query->all();
+
+        if (!$entities) {
+            return $this->_notFound($id, $subject);
+        }
+
+        $subject->set(['entities' => $entities, 'success' => true]);
+        $this->_trigger('afterFind', $subject);
+        return $entities;
+    }
+
+    /**
      * Throw exception if a record is not found
      *
      * @param string $id Record id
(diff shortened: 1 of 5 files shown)


tests/TestCase/Controller/Component/CrudComponentTest.php
```diff
@@ -120,6 +121,16 @@ public function index()
     {
         return $this->Crud->execute('index');
     }
+
+    /**
+     * Test it should render 'list'
+     *
+     * @return void
+     */
+    public function listing()
```
ankr: Since `list` is a reserved word I had to rename this method. Makes me wonder if `ListAction` should be renamed completely.

lorenzo:
I think we should rename the action to something else, since list is a reserved word for a function in PHP. Otherwise it will become difficult to declare the action in the controller and add listeners/custom logic to it

ankr:
@lorenzo Agree, also what I mentioned in my comment above. Any good suggestions for a name?

jippi:
It seems extremely similar to lookup action ? maybe those two things could be somewhat merged into one thing ?

ankr:
`LookupAction` also uses pagination

lorenzo:
@ank `all` ?

jippi:
@ankr couldn't it just be index() where you allow it to not call paginate() as an option?

ankr:
@jippi That would cause `index` action to behave very differently based on an config setting, and also other events should be dispatched as `(before|after)Paginate` would no longer apply.

I would personally prefer a new action as in this PR - but I'm very open for discussion.

src/Traits/FindMethodTrait.php
```diff
@@ -56,6 +56,35 @@ protected function _findRecord($id, Subject $subject)
...
+     * Find all records
+     *
+     * @param \Crud\Event\Subject $subject Event subject
+     * @return \Cake\ORM\Entity
+     */
+    protected function _findAll(Subject $subject)
+    {
+        $repository = $this->_table();
+
+        $query = $repository->find($this->findMethod());
```
lorenzo: You we have an event triggered after this? So we can modify the query before it is executed.

ADmad:
I too don't like the name `List` for same reason that @lorenzo pointed out.

lorenzo:
@ankr Can you change the name to `ListAll`or something like that?

t73biz:
Typically controller actions are a singular verb. After checking with thesaurus.com, a suitable alternative to list seems to be itemize. Although this might not be the most suitable since it has an -ize suffix, it does convey what the action is intending.

jippi:
seems to be stalled

davidyell:
I'd actually be in favour of a config option.

PR #386, opened 2016-02-09
Fallback to default factory for controller

jadb:
Right now, if one defines a default factory the cake promoted way using `ModelAwareTrait::modelType()`, the plugin doesn't automatically detect it.

This change will fix the behavior and help avoiding to set the factory twice (for controller and for Crud's proxy).

### DIFF
--- a/src/Core/ProxyTrait.php
+++ b/src/Core/ProxyTrait.php
@@ -111,7 +111,7 @@ trait ProxyTrait
     protected function _table()
     {
         return $this->_controller()
-            ->loadModel(null, $this->config('modelFactory') ?: 'Table');
+            ->loadModel(null, $this->config('modelFactory') ?: $this->_controller()->modelType());
     }
 
     /**


jadb:
Ok, serious fails but having a quick look, I can't see how they are related to the change. Anyone?

jippi:
Looks like mocks gone wrong, maybe :)

jadb:
That's what I thought first but the failing call isn't even about that method call to `Controller::modelType()` I added.

jippi:
bump of love!

jadb:
I never figured the failures (digged twice and pinged others for advice). Am not sure they are related.

jadb:
@ADmad could failures here be resolved too?

ADmad:
Yup #388 takes care of those.

jadb:
@ADmad @jippi - who wants to merge then?

PR #387, opened 2016-02-11
Switched validation error code, closes #337

### LINKED ISSUE #337: Validation Exception: HTTP Code 412 vs 422
Why is the HTTP code 412 (Precondition Failed) used for a validation exception and not 422 (Unprocessable Entity)?

It seems to me that 412 is for conditional headers.
>  The 412 (Precondition Failed) status code indicates that one or more
   conditions given in the request header fields evaluated to false when
   tested on the server.  This response code allows the client to place
   preconditions on the current resource state (its current
   representations and metadata) and, thus, prevent the request method
   from being applied if the target resource is in an unexpected state.
(http://tools.ietf.org/html/rfc7232#section-4.2)

And 422 for validation errors 
> The 422 (Unprocessable Entity) status code means the server
   understands the content type of the request entity (hence a
   415(Unsupported Media Type) status code is inappropriate), and the
   syntax of the request entity is correct (thus a 400 (Bad Request)
   status code is inappropriate) but was unable to process the contained
   instructions.  For example, this error condition may occur if an XML
   request body contains well-formed (i.e., syntactically correct), but
   semantically erroneous, XML instructions.
(http://tools.ietf.org/html/rfc4918#section-11.2)

jadb:
This will need to wait for cakephp/cakephp#8237 before being merged.

Note: there will be unrelated errors that are happening on other PRs too. Tried quickly resolving, no luck. Any pointers welcomed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.4
   - 5.5
   - 5.6
   - 7.0
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         }
     ],
     "require":{
-        "cakephp/cakephp": "~3.1"
+        "cakephp/cakephp": "^3.2.2"
     },
     "require-dev": {
         "phpunit/phpunit": "*",
--- a/src/Error/Exception/ValidationException.php
+++ b/src/Error/Exception/ValidationException.php
@@ -33,7 +33,7 @@ class ValidationException extends BadRequestException
      * @param \Cake\ORM\Entity $entity Entity
      * @param int $code code to report to client
      */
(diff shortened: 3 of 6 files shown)


jadb:
The related PR was merged on CakePHP.

lorenzo:
Let's wait for a release containing that commit

jippi:
should also bump composer.json to require that cakephp release version

ADmad:
@jippi Done
@jadb You are welcome :stuck_out_tongue:

jadb:
@ADmad thanks! they weren't related, were they?

ADmad:
The fails were unrelated.

PR #390, opened 2016-03-07
idField should be in list of columns

steefaan:
Before:

Call on `/products/lookup.json?id=3` prdocued:

image

After:

Call on `/products/lookup.json?id=3` produces:

image

### DIFF
--- a/src/Action/LookupAction.php
+++ b/src/Action/LookupAction.php
@@ -56,7 +56,8 @@ class LookupAction extends BaseAction
         $request = $this->_request();
 
         $config = (array)$this->config('findConfig');
-        if ($idField = $request->query('id')) {
+
+        if (($idField = $request->query('id')) && in_array($idField, $this->_table()->schema()->columns())) {
             $config['keyField'] = $idField;
         }
 


josegonzalez:
Note: won't be a bc-compatible change.

steefaan:
I use a CORS lookup with select2. In my database I save the ID of the related record but I will loose the name of the record. When I open the edit page later again, only the ID is visible but I want to show the name of the record instead, because no one can remember ID's :)

steefaan:
@josegonzalez Not sure about the breaking change here. In general yes you're right but on the other side why someone should have changed the `idField`, when this field doesn't exist in the schema?

josegonzalez:
Heh, your reasoning is almost exactly why I wrote my own AutocompleteAction.

I say its bc breaking because it turns from an array to a dictionary.

lorenzo:
I think the change makes sense. The only purpose of this action is to do autocompletes, and the feature would be broken without this change

src/Action/LookupAction.php
```diff
@@ -56,7 +56,8 @@ protected function _getFindConfig()
         $request = $this->_request();
 
         $config = (array)$this->config('findConfig');
-        if ($idField = $request->query('id')) {
+
+        if (($idField = $request->query('id')) && in_array($idField, $this->_table()->schema()->columns())) {
```
lorenzo: Using assignment inside the `if` is confusing to me

steefaan: Agree I also don't like it. I will adjust it. Additionally I would do the same conditions for `valueField`, what do you think?

lorenzo: Agree

PR #392, opened 2016-03-12
Now possible to edit the query in view action via beforeFind()

Phillaf:
I'm a little confused with tests in this repo. Any pointers as to where / how to test this case?

### DIFF
--- a/src/Traits/FindMethodTrait.php
+++ b/src/Traits/FindMethodTrait.php
@@ -44,7 +44,7 @@ trait FindMethodTrait
         ]);
 
         $this->_trigger('beforeFind', $subject);
-        $entity = $query->first();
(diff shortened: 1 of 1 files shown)


dakota:
You can already edit the query in the view action without this change. Php
objects are always passed by reference.

Phillaf:
I don't believe this does not cover cases where the query object is replaced by a new one. For example to get rid of the primary key condition, I would do this:

```
public function view($id = null)
{
    $this->Crud->on('beforeFind', function(\Cake\Event\Event $event) {
        $event->subject->query = $this->Profiles->find()
            ->where(['user_id' => $this->Auth->user('id')]);
    });
    $this->Crud->execute();
}
```

lorenzo:
Makes sense :)

PR #403, opened 2016-03-22
Update doc intro

thumbtech:
Clarified current find()->contain limitation vs. relatedModels functionality


davidyell:
I don't agree with this change. The default behaviour for add and edit actions is to create lists for related data fields, which does make them compatible with baked views.

Related data will not be loaded for index actions as listed in the documentation.

ADmad:
I agree with @davidyell this update is misleading.

thumbtech:
Learned how custom actions with beforeFind / beforePaginate events support contains! Thank you!

PR #406, opened 2016-03-23
Resolves #393

### LINKED ISSUE #393: [Docs] Prefix routing
Need some examples of using Crud just inside a prefix. In my project that I'm using as reference I've just created a new controller, which extends `AppController` and then configured it in the child controller.

If it's possible to pass routing prefixes into the configuration then we should document that. Such as just running Crud on `admin` prefixes only.

davidyell:
Added a new section to the Configuration page to give an example of how to enable Crud for just a single routing prefix.

Also fixed some minor indenting.


lorenzo:
Thanks!

PR #411, opened 2016-03-23
lookup fields should be inside schema

steefaan:
@lorenzo Sorry totally forgot to adjust the other PR.

refs https://github.com/FriendsOfCake/crud/pull/390


lorenzo:
thanks :)

PR #417, opened 2016-04-21
Added doc on how to load alternate model

LoveMyData:


steefaan:
Not sure if this should be a part of the CRUD docs. It's the common cake way.

LoveMyData:
I thought it will be a nice to have as in cake, I normally use $this->loadModel();

steefaan:
`$this->loadModel()` and `$this->modelClass = 'TableName';` do different things and isn't comparable IMO.

LoveMyData:
sorry, i am new to cake, just trying my best to help, I am still learning what are the differences. Thanks for pointing this out, this make me 'think' more now.

ADmad:
I agree with @steefaan this is basic CakePHP, not specific to Crud.

ADmad:
@LoveMyData Out of curiosity why is your default model name different than your controller name?

LoveMyData:
I am writing an API, I want the API be more friendly name but my table is more for me to remember. I think it is also to do with hosting and prefix, can't remember what is the exact reason, but I ended up 'SQL' table different to Controller name

ADmad:
Not having controller and table name as per conventions will cause unnecessary extra work. You should use routes if need different URLs then what you get using controller name.

PR #421, opened 2016-05-12
Remove .semver file.

ADmad:
It's an unwanted chore updating it on each release which
is almost always forgotten.

### DIFF
--- a/.semver
+++ b/.semver
@@ -1,5 +1,0 @@
----
-:major: 4
-:minor: 2
-:patch: 0
-:special: ''


bravo-kernel:
:+1:

jippi:
@ADmad my release tool uses the file :P i should just share the file :)

josegonzalez:
Drop it from your release tool and use tags instead.

ADmad:
@jippi My release tool is github's UI :stuck_out_tongue:

PR #425, opened 2016-05-30
remove unnecessary renderAs()

steefaan:
I need to render templates for specific API actions. Therefore I set `_serialize` to false in `Crud.beforeRender`. If I keep the deleted lines, it will search for templates inside `Posts/json/json/view.ctp` instead of `Posts/json/view.ctp`.

### DIFF
--- a/src/Listener/ApiListener.php
+++ b/src/Listener/ApiListener.php
@@ -193,13 +193,7 @@ class ApiListener extends BaseListener
         $this->_ensureData($subject);
         $this->_ensureSerialize();
 
-        $controller = $this->_controller();
-
-        if (!empty($controller->RequestHandler->ext)) {
-            $controller->RequestHandler->renderAs($controller, $controller->RequestHandler->ext);
-        }
-
-        return $controller->render();
+        return $this->_controller()->render();
     }
 
     /**


jippi:
can you please share your app code? i don't have this problem in my app :)

steefaan:
It's nothing special at all. I have an `Api` route prefix with a corresponding `AppController`. Inside my `AppController` I configure CRUD and load `Crud.Api` listener. Then I have a `PostsController` which extends from the `AppController`, where I do the following to disable automatic serialization:

```
    public function index()
    {
        $this->Crud->on('beforeRender', function (Event $event) {
            $this->Crud->controller()->set('_serialize', false);
        });

    }
```

steefaan:
Inside my `AppController` I also load `RequestHandler` with default configuration, so simply: `$this->loadComponent('RequestHandler');`

steefaan:
I also extended the `JsonView` like.

```
<?php

namespace App\View;

use Cake\View\JsonView as BaseJsonView;

class JsonView extends BaseJsonView
{
    public function initialize()
    {
        parent::initialize();

        // ... loading helpers
    }
}
```

If I put `public $subDir = '';` in here, it works as well but looks wrong.

ADmad:
Looks reasonable to me. RequestHandler already called `renderAs()` if `$ext` is not empty so doing so in action too does seem redundant.

steefaan:
@ADmad What do you think about the code coverage? It's a decrease by 29%.

ADmad:
Codecov has gone bonkers recently. Ignore it.

steefaan:
@jippi could you merge this and maybe create a release tag? I really need this, I tried to workaround the `json/json` issue but there is no clean solution.

ADmad:
@steefaan You can just use your branch through composer for now.

steefaan:
@ADmad I know but this would be additional work for a bug which is already fixed. If a release tag makes more work for you I'm also fine with a merge only :)

steefaan:
Any updates here?

PR #427, opened 2016-06-08
Fix resolving ambiguous field names in SQL searches.

thinkingmedia:
I ran into a problem where my table had an association which had a duplicate column name from the parent. Which raises a MySQL error that the column in the where clause was ambiguous.

In my `DocumentsTable` class I had to use the full name for the field.

```
   $this->searchManager()->value('Documents.active');
```

This produced a query string like this `documents?Documents[active]=0` which gets resolved as a nested array in the `$this->request->query`

So I modified `SearchListener` to flatten the query parameters back to dot notation.

This PR includes that change.

### DIFF
--- a/src/Listener/SearchListener.php
+++ b/src/Listener/SearchListener.php
@@ -3,6 +3,7 @@ namespace Crud\Listener;
 
 use Cake\Core\Plugin;
 use Cake\Event\Event;
+use Cake\Utility\Hash;
 use RuntimeException;
 
 class SearchListener extends BaseListener
@@ -60,7 +61,7 @@ class SearchListener extends BaseListener
             ));
         }
 
-        $filterParams = $table->filterParams($this->_request()->query);
+        $filterParams = $table->filterParams(Hash::flatten($this->_request()->query));
         $event->subject->query->find('search', $filterParams);
     }
 }


ADmad:
You won't have ambiguous column errors if you escape the field in filter config `$this->searchManager()->value('active', ['field' => $this->aliasField('active')]);`

You change will require everyone to change their search filter config which isn't practical.

davidyell:
Always use `aliasField()` everywhere is fast becoming a rule of thumb for me!

ADmad:
Hmm.. Maybe we can always alias the field internally in the Search manager itself.

thinkingmedia:
That needs to get into the docs.

src/Listener/SearchListener.php
```diff
@@ -60,7 +61,7 @@ public function injectSearch(Event $event)
             ));
         }
 
-        $filterParams = $table->filterParams($this->_request()->query);
+        $filterParams = $table->filterParams(Hash::flatten($this->_request()->query));
```
jippi: why flatten it?

ADmad:
> That needs to get into the docs.

A PR updating the docs is welcome :)

Closing as there is already a way to avoid this issue and proposed change would cause problems for existing users.

PR #429, opened 2016-06-17
Add an example on how to change API return data

dakota:
It's possible to customise the data that is returned with a add action for the API listener. However, how to do this is not currently documented. This documents it.


jippi:
Thanks

Issue #431, opened 2016-06-28
FindMethodTraid::_findRecord() don't works if table have composite keys.


alysson-azevedo:
In the [line 39](https://github.com/FriendsOfCake/crud/blob/master/src/Traits/FindMethodTrait.php#L39), the `current($query->aliasField($repository->primaryKey()))` returns an array.

Here's the error:
```
Warning (2): strpos() expects parameter 1 to be string, array given [CORE/src/Datasource/QueryTrait.php, line 210]
Warning (2): explode() expects parameter 2 to be string, array given [CORE/src/Datasource/QueryTrait.php, line 214]
Warning (2): Illegal offset type [ROOT/vendor/friendsofcake/crud/src/Traits/FindMethodTrait.php, line 39]
```

Is there a way to handle it?

lorenzo:
No, there is not. In cases like this I have done my own Action class.

You are also welcome to submit a pull request for fixing your issue

alysson-azevedo:
Well, i also created my own action class, but if change the FindMethodTrait->findRecord() to do the same thing Table->get() does, it will be nice.

I'll prepare a pull-request.

alysson-azevedo:
Pull-request open.

PR #432, opened 2016-06-28
Handles composite primary keys in _findRecord.

alysson-azevedo:
With this change, when we have a table with composite primary keys, all we need to do is pass the keys to the execute method.

``` php
public function view($id)
{
    $primaryKeys = explode(';',$id);
    return $this->Crud->execute(null, $primaryKeys);
}
```

Refs: #431.

### DIFF
--- a/src/Controller/Component/CrudComponent.php
+++ b/src/Controller/Component/CrudComponent.php
@@ -218,7 +218,9 @@ class CrudComponent extends Component
         $this->_action = $controllerAction ?: $this->_action;
 
         $action = $this->_action;
-        $args = (!empty($args)) ? [$args] : $this->_request->params['pass'];
+        if (empty($args)) {
+            $args = $this->_request->params['pass'];
+        }
 
         try {
             $event = $this->trigger('beforeHandle', $this->getSubject(compact('args', 'action')));
--- a/src/Controller/Component/CrudComponent.php
+++ b/src/Controller/Component/CrudComponent.php
@@ -218,9 +218,7 @@ class CrudComponent extends Component
         $this->_action = $controllerAction ?: $this->_action;
 
         $action = $this->_action;
-        if (empty($args)) {
-            $args = $this->_request->params['pass'];
-        }
+        $args = (!empty($args)) ? [$args] : $this->_request->params['pass'];
 
         try {
             $event = $this->trigger('beforeHandle', $this->getSubject(compact('args', 'action')));
--- a/src/Traits/FindMethodTrait.php
+++ b/src/Traits/FindMethodTrait.php
@@ -1,8 +1,8 @@
 <?php
 namespace Crud\Traits;
 
-use Crud\Event\Subject;
 use Cake\Datasource\Exception\InvalidPrimaryKeyException;
+use Crud\Event\Subject;
 
 trait FindMethodTrait
 {
--- a/src/Traits/FindMethodTrait.php
+++ b/src/Traits/FindMethodTrait.php
@@ -2,6 +2,7 @@
 namespace Crud\Traits;
 
 use Crud\Event\Subject;
+use Cake\Datasource\Exception\InvalidPrimaryKeyException;
 
 trait FindMethodTrait
 {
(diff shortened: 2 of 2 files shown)


src/Controller/Component/CrudComponent.php
```diff
@@ -218,9 +218,7 @@ public function execute($controllerAction = null, $args = [])
         $this->_action = $controllerAction ?: $this->_action;
 
         $action = $this->_action;
-        if (empty($args)) {
-            $args = $this->_request->params['pass'];
-        }
+        $args = (!empty($args)) ? [$args] : $this->_request->params['pass'];
```
davidyell: Urgh, shorthand. Dirty.

davidyell:
Casting non-composite keys to an array seems like overkill here. I'd be far more tempted to just check if it's composite and pass it off to a different method. Otherwise the code looks overly complex for most use-cases.

Visually I'd prefer the edge-case to be handled in it's own method.

alysson-azevedo:
I wasn't [my idea](https://github.com/alysson-azevedo/cakephp/blob/master/src/ORM/Table.php#L1164-L1182).

src/Controller/Component/CrudComponent.php
alysson-azevedo: This `[$args]` is something i wasn't sure about.
When i call `$this->Crud->execute(null, ['key1', 'key2'])` without the `[$args]`, the `BaseAction->handle()` do `return call_user_func_array([$this, $method], $args)`, then only my "key1" will go to the `_findMethod()`.

In this case, or i use that `$args = [$args]`, or user will have to use  

    $this->Crud->execute(null, [["key1","key2"]]);

src/Traits/FindMethodTrait.php
```diff
@@ -30,13 +31,34 @@ public function findMethod($method = null)
...
 
+        $key = (array)$repository->primaryKey();
+        $alias = $repository->alias();
+        foreach ($key as $index => $keyname) {
+            $key[$index] = $alias . '.' . $keyname;
+        }
+        $primaryKey = (array)$id;
+        if (count($key) !== count($primaryKey)) {
+            $primaryKey = $primaryKey ?: [null];
+            $primaryKey = array_map(function ($key) {
```
ADmad: What is this supposed to do?

src/Controller/Component/CrudComponent.php
ADmad: This isn't right, you should be using `[["key1","key2"]]` when calling the method not mucking up the args inside `execute()`. This change will break all existing actions which take multiple args.

src/Traits/FindMethodTrait.php
alysson-azevedo: I just used the same logic as `Table->get()`: https://github.com/alysson-azevedo/cakephp/blob/master/src/ORM/Table.php#L1164-L1182

That array map just adds a ' in each value of the array. A `["key"]` becomes `["'key'"]`.

src/Controller/Component/CrudComponent.php
alysson-azevedo: Ok, i'll revert that.

ADmad:
Tests for the new functionality would be nice.

alysson-azevedo:
I'd love to do that, but to be honest i don't have the knowledge needed.
I know how phpunit works and normally i'm able to create tests following examples from tutorials or other existing tests (and adapting them).

But i couldn't find any tests here that i could use as example to test. The Blogs and Users don't have composite keys and i don't know what i need to do in this case.

But if anyone help me, i can try.

ADmad:
You can use the test cases of the same method you got the code from ;)   Core fixtures are also available, you can include them using `'core.foo` in `$fixtures`.

alysson-azevedo:
Well, i couldn't figure out how to implement the tests.
I know it is trivial, but for me it is really confusing.

If this is a problem, feel free to close this pull, all i did was copy the cakephp's code anyways.

src/Traits/FindMethodTrait.php
```diff
@@ -30,13 +31,34 @@ public function findMethod($method = null)
...
+     *      incorrect number of elements.
      */
     protected function _findRecord($id, Subject $subject)
     {
         $repository = $this->_table();
 
+        $key = (array)$repository->primaryKey();
+        $alias = $repository->alias();
+        foreach ($key as $index => $keyname) {
+            $key[$index] = $alias . '.' . $keyname;
```
jippi: here you can use `$repository->aliasField()` instead

jippi:
@lorenzo can you give this a look when you got time - the code is copy from the core, but it seem kinda over-complicated

ADmad:
This can be merged once tests are provided. The new code added currently doesn't have any test coverage.

dereuromark:
Should the core provide a method for the "id to conditions" part ([here](https://github.com/alysson-azevedo/cakephp/blob/master/src/ORM/Table.php#L1164-L1182).)?
That way it would be more DRY to reuse the core method doing this.

jippi:
if we merge this, all view,add,edit,delete need to support composite keys imo - though the PR seems stalled :)

alysson-azevedo:
i'm sorry everone. I proposed this PR because the code i used was the same used on cake3. But since it is not good, i think we can close this.

For those who need to work with composite keys, it's easy to override the `_findRecord` method anyways.

alysson-azevedo:
Cool. I'll close this pr since there's a better solution.

PR #437, opened 2016-07-26
Fixed name of custom action in documentation

drmonkeyninja:
The name of the action was different to the `className` used in the example usage further down the page.


ADmad:
Thanks

Issue #310, opened 2015-08-03
jsonapi.org 1.0 compability?


Suven:
Hey all,

What do you think about jsonapi.org-compability?

I like the fact that it is at least some 'common' standard which different backend- and frontend-applications may implement. One could argue about how pragmatic or feasible that particular standard is when you are only building your 'own' API.

I definitely would like to work on smth cake-ready that implements it (maybe via https://github.com/tobscure/json-api ). The question is wether or not you are interested in seeing it in CRUD or if I am creating a plugin from scratch.

ADmad:
> ...if I am creating a plugin from scratch.

You don't need to start from scratch. Just create a plugin which contains listeners which can be used instead of Crud's ApiListener and ApiPaginationListener.

jippi:
@Suven by all means, submit it as a listener for CRUD ! :)

jadb:
I think the having the option to switch between preferred "standards" would be neat. Go for it @Suven!

michelpeters:
So, do I understand it correctly when saying that this will also include support for URL's like used on the referenced jsonapi.org website: `http://example.com/posts/1/comments`? Because, this is not how the normal Cake URL conventions work. I know you can write your own router code, but I guess it gets complex pretty quickly (and irritating to add all the urls in all different incarnations ).

I've read the spec on jsonapi.org quickly, so I might have missed it, but I'm not sure the URL format is part of the spec. If it is, I'm certainly interested in this project as well. If it's not, ofcourse I still applaud more specs/formats to be supported.

jadb:
Cake works this way already, it's just not advised to nest a lot of resources. One level is not so bad though.

[reference](http://book.cakephp.org/3.0/en/development/routing.html#creating-nested-resource-routes)

jippi:
I'm fine with both - don't see the hard requirement to nest the resources, can be flat by default and then nested by simply changing your routes later per app

ADmad:
Closing as no one has shown interest in implementing this. Those looking to generate responses following jsonapi standard can check out https://github.com/josbeir/cakephp-json-api

bravo-kernel:
Fellow friends, I have opened a PR containing a JsonApiListener, all feedback more than welcome.

Issue #416, opened 2016-04-14
API bulk actions returning true even if some records aren't updated


aceat64:
When bulk actions are performed if any rows are updated then `success` is true, if no rules are updated `success` is false.

I think it would make sense for `success` to be false if only some of the records were updated, and then have `data` provide a message stating that the update was partially successful and showing the count of how many records were updated.

Ideally we'd show the ids of which records weren't updated, but I think that would require a 2nd query and some overhead.

Thoughts?

lorenzo:
Maybe it should be a mode flag? I think I prefer the current default

davidyell:
To me, the ideal situation would it returning an array of errors which data which failed to update instead of false, this would allow the developer to address the fail condition in a more concise way.

josegonzalez:
Going to reopen this. I believe it should be done, someone just has to sit down and write the code.

bravo-kernel:
I could certainly use this feature for singe `add` as well as this would complement the jsonapi spec.

PR #445, opened 2016-09-11
JsonApiListener

bravo-kernel:
Refs #310. This is a working prototype of the [JsonApi](http://jsonapi.org) listener. It uses [neomerx/json-api](https://github.com/neomerx/json-api), a lot of inspiration from [josbeir/cakephp-json-api](https://github.com/josbeir/cakephp-json-api) and currently supports `index`, `view`, `add`, `edit`, `delete` and `error` responses.

The reason I'm putting it out here is that I believe this would be a most valuable listener to Crud and CakePHP as it would make our framework instantly compatible with a lot of tools and frameworks out there like Ember Data. Given the fact that version 1.1 of the jsonapi spec is coming this might also be the right time.

### GENERAL
- [x] warn user if suggested composer package is not installed
- [X] require the  `application/vnd.api+json` Accept request header (by adding detector)
- [X] require the  `application/vnd.api+json` Content-Type request header when posting data
- [x] support all [php predefined json constants](http://php.net/manual/en/json.constants.php)
- [x] remove `_id` fields from response body as [described here](http://jsonapi.org/format/#document-resource-object-attributes)
- [x] add custom route with support for (belongsTo) relationship `self` links
- [X] ~extend custom route to support (hasMany) relationship `self` links~
- [X] generate `self::SHOW_SELF` relationship-self-belongsTo-link: `/countries/1/relationships/currency`
- [X] generate `self::SHOW_SELF` relationship-self-hasMany-link: `/cultures?country_id=1`
- [X] ApiPaginationListener support
- [X] ApiQueryLogListener support
- [X] add tests
- [X] prevent redirects after delete, add, etc. (by implementing beforeRedirect stop event)
- [ ] document

### PRESENTING RESOURCES (READ)
- [X] automatically generate json api for single entity using dynamic schema
- [X] automatically generate json api for entity collections using dynamic schema
- [X] allow overrides by creating manual schema(s)
- [X] support resource-less responses (empty body or [meta node only](https://github.com/neomerx/json-api/wiki#encodemeta-method))
- [X] respond with `application/vnd.api+json` Content-Type response header
- [X] set viewVars using Crud listener configuration options
- [x] properly add top-level `query` node to 200 responses (respecting `jsonOptions` config option)
- [x] implement neomerx exception renderer for default errors

## CREATING RESOURCES (CREATE/UPDATE)
- [x] create resources without relationships
- [x] create resources with relationships (belongsTo only as per the spec)
- [x] update resources without relationships
- [x] update resource along with belongsTo relationships
- [X] ~update resource along with hasMany relationships (unsure if the spec supports this atm)~
- [x] add `DocumentValidator` to ensure incoming request data structure respects the [jsonapi crud specification](http://jsonapi.org/format/#crud)
- [x] properly decode incoming jsonapi data to cake compatible format
- [x] implement neomerx exception renderer for (both entity and document) validation errors
- [x] enforce using `PATCH` instead of `PUT` as [described here](http://jsonapi.org/faq/#wheres-put)
- [x] display correct body after successful `add` action (with and without relationships)
- [x] **fix DocumentValidator bug**: correctly posted hasMany relationship data is now being flagged invalid (claiming `type` and `id` nodes are missing when they actually exist).

## DELETING RESOURCES
- [x] respond with 204 status code, empty body and no redirect after successful delete

### NICE TO HAVES:
- [X] make JSON_PRETTY_PRINT optional for debug mode
- [ ] support (JSON API spec) optional `related` [links](http://jsonapi.org/format/#document-links)
- [ ] configurable [links](http://jsonapi.org/format/#document-links) in error responses (now hardcoded)
- [ ] respect `jsonOptions` config option when generating exceptions. Will probably require a custom ErrorController to be placed in the app by end-user
- [ ] support status codes 415 and 406 as [described here](http://jsonapi.org/format/#content-negotiation-servers)
- [ ] automatic `success` member a-la ApiListener inside top-level `meta` member as config option
- [ ] ability to respond with 200 status code and `meta` (only) body after successful delete ([as described here](http://jsonapi.org/format/#crud-deleting))
- [ ] ~on success send optional `Location` header with link to created resource as [described here](http://jsonapi.org/format/#crud-creating-responses-201).~ Removed from prototype because it caused a CORS redirect error
- [ ] jsonp?

### MISC:
- [x] solved: why are NeoMarx options not respected (e.g. show-self-top-level)
- [x] solved: fix missing SELF link in `included` member ?

ADDITIONAL READING:
- [discuss.jsonapi.org](http://discuss.jsonapi.org)
- [jsonapi limitations when creating records with relationships](https://emberigniter.com/saving-models-relationships-json-api/)


lorenzo:
I like the idea @bravo-kernel

bravo-kernel:
I have updated all sources for which I think is now reviewable code. This gives me a pretty good idea of the size and complexity, not as bad as I thought beforehand.

All feedback more than welcome. Enabling the listener should be all that is needed to test drive.

### Posting data

I have deliberately kept [decoding incoming jsonapi data](https://github.com/FriendsOfCake/crud/pull/445/files#diff-62d4f5f8797dc1fd34406920bd2a38baR246) as minimal/simple as possible atm because I am not sure if this is the desired approach.

To test posting data use the following jsonapi body format when creating a new record without foreign keys:

```json
{
  "data": {
    "type": "countries",
    "attributes": {
      "code": "FR",
      "name": "France",
      "native": "République française"
    }
  }
}
```

To create a new record with foreign key use something similar to:

```json
{
  "data": {
      "type": "countries",
      "id": "1",
      "attributes": {
        "code": "NL",
        "name": "The Netherlands",
        "native": "Nederland",
        "currency_id": 1
      },
      "relationships": {
        "currency": {
          "data": {
            "type": "currencies",
            "id": "2"
          }
        },
        "cultures": {
          "data": [
            {
              "type": "cultures",
              "id": "3"
            }
          ]
        }
      }
  }
}
```

src/Error/JsonApiExceptionRenderer.php
```diff
@@ -0,0 +1,205 @@
+<?php
+namespace Crud\Error;
+
+use Cake\Core\Configure;
+use Cake\Datasource\ConnectionManager;
+use Cake\Error\Debugger;
+use \Neomerx\JsonApi\Document\Error;
```
jippi: don't need the `\`

src/Controller/Component/CrudComponent.php
```diff
@@ -649,7 +649,13 @@ protected function _loadListener($name)
                 throw new ListenerNotConfiguredException(sprintf('Listener "%s" is not configured', $name));
             }
 
+            // @todo: temporary hack to prevent `undefined index: className` error because `api` is in listeners list even though not enabled.
```
jippi: how does this case happen?

src/Listener/JsonApiListener.php
```diff
@@ -0,0 +1,281 @@
...
+{
+
+    /**
+     * Default configuration
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'detectors' => [
+            'jsonapi' => ['ext' => 'json', 'accepts' => 'application/vnd.api+json'],
```
jippi: if `api`listener is loaded too, the load order will decide which listener that will respond?

src/Listener/JsonApiListener.php
```diff
@@ -0,0 +1,281 @@
...
+    public function render(Subject $subject)
+    {
+        $controller = $this->_controller();
+        $controller->viewBuilder()->className('Crud.JsonApi');
+
+        // Validate configuration before creating corresponding viewVars
+        if ($this->config('urlPrefix') !== null && !is_string($this->config('urlPrefix'))) {
+            throw new CrudException('JsonApiListener configuration option `urlPrefix` only accepts a string');
+        }
+        $controller->set([
```
jippi: instead of calling `set()` over and over, maybe buffer them in an array and call `set` with that?

src/Listener/JsonApiListener.php
```diff
@@ -0,0 +1,281 @@
...
+        // Set required viewVar with ORM AssociationCollection for the current
+        // model so it can be used to generate the `relationships` nodes in the
+        // response.
+        //
+        // Please note that we are removing associated models not found in
+        // the find() result to prevent `null` relationships appearing in
+        // the response.
+        $controllerName = $controller->name; // e.g. Countries
+        $entityName = Inflector::singularize($controllerName); // e.g. Country
+        $table = $controller->$controllerName; // table object
```
jippi: `$controller->${controllerName}` 

isn't it really a `$tableName` ? :)

src/Listener/JsonApiListener.php
```diff
@@ -0,0 +1,281 @@
...
+
+        return $controller->render();
+    }
+
+    /**
+     * Helper method to get query log.
+     *
+     * @return array Query log.
+     */
+    protected function _getQueryLog()
```
jippi: this seem duplicated now, mabe move the common chunks into a trait or an abstract shared class :)

src/View/todo.md
```diff
@@ -0,0 +1,35 @@
...
+- [X] add detector for `application/vnd.api+json` Accept request header 
+- [X] set viewVars using Crud listener configuration options
+- [X] respond with `application/vnd.api+json` Content-Type response header
+- [X] suggest neomerx composer package instead of require
+- [X] implement neomerx exception renderer for default errors
+- [X] implement neomerx exception renderer for validation errors
+- [X] add top-level `debug` and `query` nodes to exceptions
+- [X] add top-level `query` node to 200 responses
+- [ ] describe required custom route to support incoming requests to `self::SHOW_RELATED`
+- [ ] decode incoming jsonapi data to cake compatible format
```
jippi: done?

jippi: also, shouldn't this file be somewhere else? :)

src/Error/JsonApiExceptionRenderer.php
bravo-kernel: Fixed in https://github.com/FriendsOfCake/crud/pull/445/commits/6a70b299a063ead55c934455f5fb491c811426bb

src/Listener/JsonApiListener.php
bravo-kernel: Good point, fixed in https://github.com/FriendsOfCake/crud/pull/445/commits/6a70b299a063ead55c934455f5fb491c811426bb by adding `validateConfigOptions()` method found [here](https://github.com/FriendsOfCake/crud/pull/445/files#diff-62d4f5f8797dc1fd34406920bd2a38baR164)

src/Listener/JsonApiListener.php
bravo-kernel: Correct, changed as suggested in https://github.com/FriendsOfCake/crud/pull/445/commits/6a70b299a063ead55c934455f5fb491c811426bb

src/Controller/Component/CrudComponent.php
bravo-kernel: This happens as soon as I add the JsonApiListener, regardless by itself or along with ApiListener, PaginationListener or any other combination. Somehow `api` always ends up in the list of listeners even though I did not enable it.

```
Could not find listener class:
Crud\Error\Exception\MissingListenerException

Notice (8): Undefined index: className [ROOT/vendor/friendsofcake/crud/src/Controller/Component/CrudComponent.php, line 657]
Code Context
//            }

$className = App::classname($config['className'], 'Listener', 'Listener');
Crud\Controller\Component\CrudComponent::_loadListener() - ROOT/vendor/friendsofcake/crud/src/Controller/Component/CrudComponent.php, line 657
Crud\Controller\Component\CrudComponent::_loadListeners() - ROOT/vendor/friendsofcake/crud/src/Controller/Component/CrudComponent.php, line 631
Crud\Controller\Component\CrudComponent::beforeFilter() - ROOT/vendor/friendsofcake/crud/src/Controller/Component/CrudComponent.php, line 190
Cake\Event\EventManager::_callListener() - CORE/src/Event/EventManager.php, line 422
Cake\Event\EventManager::dispatch() - CORE/src/Event/EventManager.php, line 391
Cake\Controller\Controller::dispatchEvent() - CORE/src/Event/EventDispatcherTrait.php, line 78
Cake\Controller\Controller::startupProcess() - CORE/src/Controller/Controller.php, line 497
Cake\Http\ActionDispatcher::_invoke() - CORE/src/Http/ActionDispatcher.php, line 117
Cake\Http\ActionDispatcher::dispatch() - CORE/src/Http/ActionDispatcher.php, line 96
Cake\Routing\Dispatcher::dispatch() - CORE/src/Routing/Dispatcher.php, line 60
[main] - ROOT/webroot/index.php, line 37
```

src/View/todo.md
bravo-kernel: - renamed to `properly decode incoming jsonapi data` ;)
- personal todo list deleted from the repo

src/Listener/JsonApiListener.php
bravo-kernel: You are totally right, I did not want to touch (or depend on) anything existing but your suggestion solves all that so I I just moved this into a trait in https://github.com/FriendsOfCake/crud/pull/445/files#diff-62d4f5f8797dc1fd34406920bd2a38baR164. Thanks, much obliged 💃

src/Listener/JsonApiListener.php
bravo-kernel: Honestly... I just took this from the ApiListener and made it work without ever giving both Listeners enabled at the same time a thought 😱. I guess a test would be useful here once I get to that point?

src/Error/JsonApiExceptionRenderer.php
```diff
@@ -16,6 +16,8 @@
  */
 class JsonApiExceptionRenderer extends \Cake\Error\ExceptionRenderer
 {
+    use \Crud\Traits\QueryLogTrait;
```
jippi: move to `use` in class and just use `QueryLogTrait` here :)

src/Listener/JsonApiListener.php
jippi: it's probably a weird edge-case :p

src/Controller/Component/CrudComponent.php
jippi: thats odd :o

src/Error/JsonApiExceptionRenderer.php
bravo-kernel: Just fixed in https://github.com/FriendsOfCake/crud/pull/445/commits/eaaf26ca9bb2e2469486a12bbce3e25d5a55e9c3, thanks (again)

src/Controller/Component/CrudComponent.php
bravo-kernel: Not odd at all, it was actually something very silly on my part again caused by an an old config setting for the ApiListener floating around in my code. Removing it solves all my problems. Undoing the hack right now.

```php
$this->Crud->config(['listeners.api.exceptionRenderer' => 'App\Error\ExceptionRenderer']);
```

bravo-kernel:
Thanks for the review @jippi, much obliged, I will start working on the tests.

The big question remains if this listener stands a chance of being adopted into this repo? In all honesty I  can empathize with both the yes (more eyes and hands, better quality) and the no (who needs more maintenance anyway). If it turns out to become a no I will follow @ADmad's advice and move it into an external plugin.

bravo-kernel:
Not sure why the existing phpunit tests are failing on Travis btw. No existing Crud was harmed IMO.

jippi:
i don't see a reason not to include it in CRUD directly

if there is a majority for having it in a separate repo, we can just create `foc/crud-jsonapi` and move it to there and add is as `suggested` in composer.json

i'm fine with both :)

ADmad:
> i don't see a reason not to include it in CRUD directly

:+1:

ADmad:
@bravo-kernel That failure exists on master, it's unrelated to your changes.

ceeram:
The listener requires neomerx/json-api as dependency, but its only in suggest in the composer.json, since its not required for other parts of crud. Having this as separate package on foc, allows to add neomerx/json-api as required dependency there. Ofcourse it can be documented that it is required to be installed when using jsonapi listener, but it might confuse users getting unexpected errors about missing classes.

Other then that im happy to see a json-api listener to arrive in foc one way or another. 🍰

josegonzalez:
👍 keep it in friendsofcake/crud. It's a single dependency, we can refactor it later if we think we can do it better.

src/Listener/JsonApiListener.php
```diff
@@ -0,0 +1,264 @@
...
+    /**
+     * beforeHandle
+     *
+     * Called before the crud action is executed.
+     *
+     * @param \Cake\Event\Event $event Event
+     * @return void
+     */
+    public function beforeHandle(Event $event)
+    {
```
jippi: for usability, it would be good service to check here if `neomerx/json-api` is installed or not, and throw an exception in case it isn't ?

bravo-kernel: I am fine with adding that check but think adding the package as composer-required dependency with specific supported version would be the preferred solution given [there is no such thing as an optional dependency](http://php-and-symfony.matthiasnoback.nl/2014/04/theres-no-such-thing-as-an-optional-dependency/).

bravo-kernel: Package is now a required dependency, if only to make test driving this easier. Can always be changed later.

jippi: for people not using this feature, i think it should be a suggested. That the dependency is optional was the primary reason why i thought it should be included in the crud-core :)

bravo-kernel: Fair enough, I guess we can throw your suggested expection and show supported package version

bravo-kernel: Hope https://github.com/FriendsOfCake/crud/pull/445/commits/a873579df93a7d83df10f31378b43ffc2af213ce pleases everyone

src/Listener/JsonApiListener.php
```diff
@@ -197,6 +205,19 @@ protected function _validateConfigOptions()
...
     /**
+     * Make sure the neomerx/json-api composer package is installed
+     *
+     * @throws \Crud\Error\Exception\CrudException
+     * @return void
+     */
+    protected function _checkPackageDependencies()
+    {
+        if (!class_exists('\Neomerx\JsonApi\Encoder\Encoder')) {
+            throw new CrudException('JsonApiListener requires composer installing' . $this->neomerxPackage);
```
jippi: missing space after `installing`

bravo-kernel: omg

bravo-kernel: it was more related to me face palming over stupid mistakes caused by trying to squeeze in commits in between things 😱

davidyell:
I can see documentation on your todo list which is great. Looking forward to giving this a try.

I'm also 👍  on including as part of the foc/crud package, due to Crud being promoted by many people, myself included, as a quick and smart way to build apis.

tests/TestCase/Listener/JsonApiListenerTest.php
```diff
@@ -369,6 +369,47 @@ public function testCheckRequestMethodsFailContentHeader()
...
+        $this->setReflectionClassInstance($listener);
+        $result = $this->callProtectedMethod('_getFindResult', [$subject], $listener);
+        $this->assertSame($subject->entities, $result);
+
+        $controller->request->action = 'any-other-action-name';
+        $this->setReflectionClassInstance($listener);
+        $result = $this->callProtectedMethod('_getFindResult', [$subject], $listener);
+        $this->assertSame($subject->entity, $result);
+    }
+
```
jippi: newlines :P

bravo-kernel: Do you mean an empty line above assertsame?

bravo-kernel: Oh f***k, I see now 😱

bravo-kernel: Nice catch. Wonder why the cake sniffer doesn't warn me about that.

jippi:
@bravo-kernel the broken test in crud got fixed in a different PR; merge in master to get all green :D

bravo-kernel:
Package requirement check works as expected :relaxed:

davidyell:
I'm very impressed with all the work that you're putting in on this new listener @bravo-kernel Great job, and thanks! 👍  :ok_hand:

bravo-kernel:
It's getting somewhere but still far from finished. Thanks for the motivational though ☺️

bravo-kernel:
The listener tests are stuck at 99.06% because of [this test](https://github.com/FriendsOfCake/crud/pull/445/files#diff-aacabc6948b30d44f3a4f8e586447f3fR207). I cannot find a way to emulate the class/composer package not being loaded. Confirmation that this is indeed impossible to test would be nice, a solution even better. TIA.

jippi:
fuck that test :P you are testing if class_exist works, we know it works

bravo-kernel:
This took some thinking but https://github.com/FriendsOfCake/crud/pull/445/commits/64b7233b4fba19157ba06da6c1ead3b433bba857 adds a `DocumentValidator` for proper (and hopefully understandable) validation of incoming data/jsonapi documents as there seem to be requirements for those as well.

I will update the feature-list on top of this page to reflect both presenting and creating.

bravo-kernel:
OK, these last commits implement full functionality so if anybody feels like test-driving... you're welcome :)

Routes are next.

bravo-kernel:
The custom route now:

- properly handles belongsTo links (e.g. http://api.app/countries/1/relationships/currency)
- does not handle hasMany links (e.g. http://api.app/countries/1/relationships/cultures)

The latter seems to be impossible to achieve inside the route class without additional functionality like the search plugin + query/search paramater but... I am not 100% sure. Confirmation would be appreciated.

bravo-kernel:
Since the feature set is now quite complete it's time to start refactoring and finalizing the tests.

If there are any comments this would be the time.

src/Listener/JsonApiListener.php
```diff
@@ -483,6 +483,12 @@ protected function _getEntityList($entityName, $associations)
      */
     protected function _checkRequestData()
     {
+        // prevent false positives for GET requests using query parameters
+        // as those will also have request data
```
ADmad: Huh?  `request->data` is not populated for GET requests.

bravo-kernel: Just tested again and I am now certain that it is pagination which is filling the request data. Changed the comment in the code accordingly. Thanks for checking and.. apologies for driving you crazy 😆 If you disagree let me know.

```php
    public function beforeFilter(Event $event)
    {
        parent::beforeFilter($event);
        $this->paginate = [
            'sortWhitelist' => [
                'code',
                'name'
            ]
        ];
    }
``

Comment (davidyell) [inline src/Listener/JsonApiListener.php]: Could also be the `friendsofcake/search` PrgComponent perhaps? As that has this behaviour.

Comment (bravo-kernel) [inline src/Listener/JsonApiListener.php]: I'm not using that component in my test-app so I guess that's not the culprit 😉

**Comment by bravo-kernel** [comment]
Project coverage bump to +90% as a bonus 🍰

**Comment by dereuromark** [comment]
92+ coverage is quite impressive! Nice job

File: src/Action/EditAction.php [hunk 13]
```diff
@@ -140,6 +140,19 @@ protected function _post($id = null)
     }
 
     /**
+     * HTTP PATCH handler
+     *
+     * Thin proxy for _put
+     *
+     * @param mixed $id Record id
```
dereuromark: string|null

src/Action/EditAction.php
```diff
@@ -140,6 +140,19 @@ protected function _post($id = null)
...
+     * @return void|\Cake\Network\Response
```
dereuromark: Fixed that in my PR to ...|null

src/Schema/JsonApi/DynamicEntitySchema.php
```diff
@@ -0,0 +1,231 @@
...
+
+        return $relations;
+    }
+
+    /**
+     * NeoMerx override to generate Cake belongsTo and hasMany links.
+     *
+     * @param \Cake\ORM\Entity $resource Entity
+     * @param string $name Relationship name in lowercase singular or plural
+     * @param array $meta Optional array with meta information
```
dereuromark: |null

bravo-kernel: Thank you for reviewing, I will try to pick up the changes later tonight.

bravo-kernel:
To start with, all functionality I could think of has been implemented and has tests. No shortcuts were made so I am pretty confident about, and pleased with, the result. 

However, I have a list of notes mentioning the following weak-points/questions and would really appreciate your expert review/feedback on them so I can push this toward a merge-able alpha-version.

- [ ] not tested with RelatedModels plugin and not sure if that would be needed
- [ ] unsure about proper handling of CamelCased controllers
- [ ] unsure about proper handling of dashed, inflected, etc routes
- [ ] unsure about proper handling of plugins
- [ ] the link-generation logic (e.g. [here](https://github.com/FriendsOfCake/crud/pull/445/files#diff-8e39d88847400fa247f192501eb3b357R140) is quite hackish and will probably fail for cases mentioned above.
- [ ] assumption was made about generated hasMany links: user would have to use search plugin to resolve `/cultures?country_id=1` link but I am not 100% sure this will work
- [ ] will the custom route be able to detect/handle plugin routes?

There are probably a lot more things to consider, feel free to 🔫

src/Schema/JsonApi/DynamicEntitySchema.php
bravo-kernel: Just removed the fqcn, thanks for pointing it out. I am pausing the other return type changes until PR #416 settles.

dereuromark: you can now rebase.

bravo-kernel:
Pretty simple example but..... Ember Data fetching records from a Crud API automagically 🎉 💃 

image

src/Listener/JsonApiListener.php
bravo-kernel: mr. @davidyell, I completely misread your comment. Somehow my brain did not pick up the plugin. It seems you were spot on, it is indeed the PrgComponent. I am updating my comment in the code. Thanks!

davidyell:
Wow, this is cool @bravo-kernel can't wait to give it a try.

src/Listener/JsonApiListener.php
ADmad: You don't need to update comment but simply remove handling of this case. It's the dev's job to disable the querystring to data setting feature of Prg component as doing so makes no sense for an API response.

src/Listener/JsonApiListener.php
```diff
@@ -0,0 +1,583 @@
...
+     * structure for resources and if so, converts that array to CakePHP
+     * compatible format so it can be processed as usual from there.
+     *
+     * @return void
+     */
+    protected function _checkRequestData()
+    {
+        $requestMethod = $this->_controller()->request->method();
+
+        if ($requestMethod !== 'POST' && $requestMethod !== 'PATCH') {
```
bravo-kernel: I tried both and only the && gave me the desired result (method skipped for everything but POST and PATCH)

bravo-kernel:
[This commit](https://github.com/FriendsOfCake/crud/pull/445/commits/6d4cd8653b51680eaafd617412b5f8904604be99) allows stopping beforeRedirect event to both:
- prevent redirects after e.g. deleting records (which make no sense for an API call)
- allow for easily responding with the JSON API 204 error and blank body message

Done like this in my controller(s):

```php
    public function beforeRedirect(Event $event)
    {
        $event->stopPropagation();
    }

    public function afterDelete(Event $event)
    {
        $this->_controller()->response->statusCode(204);
    }
```
 Thanks to mr. @jippi for the pointers 👍

bravo-kernel:
Latest commit aligns Crud API with Ember Data for automatic `view` and `edit` actions.

image

bravo-kernel:
This last commit completes my Ember test app connecting to Crud `add` action with out-of-the-box:

- validation errors (auto-mapped to corresponding local properties)
- generating belongsTo relationships

ember-add-action-validation

bravo-kernel:
[Assistance needed](https://github.com/FriendsOfCake/crud/pull/445#issuecomment-261480840)

jippi:
@bravo-kernel what's the status of this PR, is everything functional, and nothing outside of the new listener is broken?

It gotten way way way too big, so if everything kinda-works, lets merge in and submit smaller PRs to fix the rest?

src/Listener/ApiPaginationListener.php
```diff
@@ -70,4 +78,60 @@ public function beforeRender(Event $event)
...
         $this->_action()->config('serialize.pagination', 'pagination');
     }
+
+    /**
+     * Generates pagination viewVars with JSON API compatible hyperlinks.
+     *
+     * @param array $pagination CakePHP pagination result
+     * @return array
+     */
+    protected function _getJsonApiPaginationResponse(array $pagination)
```
jippi: this seem leaky, shouldn't there be a JSONApi PaginationListener instead? :)

bravo-kernel: I had that once but deliberately chose to integrate into ApiPaginationListener as this keeps things consistent when setting up. E.g. here's how I am testrunning it. All I need to do to switch between the listeners is change `JsonApi` to `Api`.

```php
        $this->_controller->loadComponent('Crud.Crud', [
            'actions' => [
                'Crud.Index',
                'Crud.View',
                'Crud.Add',
                'Crud.Edit',
                'Crud.Delete'
            ],
            'listeners' => [
                'Crud.JsonApi',
                'Crud.ApiPagination',
                'Crud.ApiQueryLog'
            ]
        ]);
```

bravo-kernel:
@jippi the status is:

- everything I was able to test/develop against works as expected
- the parts that will probably break are [in this list](https://github.com/FriendsOfCake/crud/pull/445#issuecomment-261480840)
- it should be fully BC, I have paid a lot of attention to make sure ApiListener wasn't harmed

I agree it has become quite large but stopping was not an option ;) Feel free to merge, I can create documentation as a separate PR if need be.

jippi:
good job @bravo-kernel ! :)

bravo-kernel:
You are brave sir 👍 

I will start working on the docs, hopefully this will get enough traction to become stable.

Thanks for all the help!

jippi:
Thank **you** for a monster 3 months long PR :)

josegonzalez:
@bravo-kernel docs?

bravo-kernel:
Docs were on my wishlist for the holidays but unfortunately my free-time has been claimed 😱 so probably not until next year.

PR #446, opened 2016-09-11
Does not return and has no return value.

thinkingmedia:
Confused the IDE.


josegonzalez:
You need at least the void return right after, or the trait will continue working.

ADmad:
_notFound() throws an exception so normal won't continue either way.

thinkingmedia:
This isn't a functional change. It was either remove the return or change the phpdoc.

It was just bugging phpstorm's intellisense.

Issue #402, opened 2016-03-18
relatedModels in Crud.View?


thumbtech:
AppController includes:

    $this->loadComponent('Crud.Crud', [
        'actions' => [
            'Crud.Index',
            'Crud.Add',
            'Crud.Edit',
            'Crud.View',
            'Crud.Delete'
        ]
    ]);
    $this->Crud->addListener('relatedModels', 'Crud.RelatedModels');

I have a belongsToMany relationship that is correctly handled by Crud.Add & Crud.Edit (i.e., the related data is displayed in a multi-select list and the selected items are saved to the DB).

Now I just want to *view* the related data in the view. The baked view template is looking for the data, but Crud.View is not querying for it. This seems to be the (odd) default behavior according to doc ...

So now I am setting my AppController to:

    $this->loadComponent('Crud.Crud', [
        'actions' => [
            'Crud.Index',
            'Crud.Add',
            'Crud.Edit',
            'View' => [
                'className' => 'Crud.View',
                'relatedModels' => true
            ],
            'Crud.Delete'
        ]
    ]);
    $this->Crud->addListener('relatedModels', 'Crud.RelatedModels');

But still no related data.

This seems pretty hard vs. just baking the controller ... Or?

So if src/Model/Table/ContactsTable.php =

    class ContactsTable extends Table
    {
        public function initialize(array $config)
        {
            $this->belongsToMany('Categories');
        }
    }

then the baked controller's view action obviously includes the *contain*:

    public function view($id = null)
    {
        $contact = $this->Contacts->get($id, [
            'contain' => ['Categories']
        ]);
        ...

Should Crud.View know to do the same?

lorenzo:
related models is for sending the required data to the forms for generating select lists. Currently it has no effect on `contain` for the finder queries

thumbtech:
Thank you. That is clear regarding relatedmodels in docs, but I felt a need to confirm since elsewhere in the docs it states that actions are by default compatible with baked views. But views are correctly baked with contain logic, which the actions do not support ...

dingonv:
Could someone please update the docs that in a view action, you will not get related entities in your returned entities? I just spent well over an hour banging my head against a wall wondering why the query for the related entity was returning, but was NOT being added to my output. I finally came in here read this and facedesked...

bravo-kernel:
@dingonv feel free to PR, documentation can be found in the `docs` folder of this repo.

davidyell:
[...]

PR #448, opened 2016-09-22
Adds beforeFind to RelatedModelsListener, for View

eaxexe:
Change makes it possible to get related models for Crud.View, just like we have with Crud.Index, making it more streamlined.

refs #402

### DIFF
--- a/src/Listener/RelatedModelsListener.php
+++ b/src/Listener/RelatedModelsListener.php
@@ -25,6 +25,7 @@ class RelatedModelsListener extends BaseListener
     {
         return [
             'Crud.beforePaginate' => 'beforePaginate',
+            'Crud.beforeFind' => 'beforeFind',
             'Crud.beforeRender' => 'beforeRender',
         ];
     }
@@ -36,6 +37,27 @@ class RelatedModelsListener extends BaseListener
      * @return void
      */
     public function beforePaginate(Event $event)
+    {
+        $contained = $event->subject()->query->contain();
+        if (!empty($contained)) {
+            return;
+        }
+
+        $models = $this->models();
+        if (empty($models)) {
+            return;
+        }
+
+        $event->subject()->query->contain(array_keys($models));
+    }
+
+    /**
+     * Automatically parse and contain related table classes
+     *
+     * @param \Cake\Event\Event $event Before find event
+     * @return void
+     */
+    public function beforeFind(Event $event)
     {
         $contained = $event->subject()->query->contain();
         if (!empty($contained)) {
--- a/src/Listener/RelatedModelsListener.php
+++ b/src/Listener/RelatedModelsListener.php
@@ -31,33 +31,32 @@ class RelatedModelsListener extends BaseListener
     }
 
     /**
-     * Automatically parse and contain related table classes
+     * Handle beforePaginate
      *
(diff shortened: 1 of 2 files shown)


eaxexe:
I am aware that the checks fail - However, if you pull crud and run phpunit currently, you will get the failed test, that also fails here.

The test I have added works, when I run it :)

ADmad:
This change will overlaps with [ViewListener::beforeFind()](https://github.com/FriendsOfCake/crud-view/blob/master/src/Listener/ViewListener.php#L29) which does similar thing but provides more control over the models to contain.

If this `RelatedModelsListener` runs before `ViewListener` one will lose the customizability the latter provides (since it bails out if containments already already set) and cause problems with existing app.

eaxexe:
It seems to me, that the functionality should be in crud rather than crud-view, since many projects, like mine, may want to use crud without crud-view.

Beyond that, it is a streamlining of relatedModels<http://crud.readthedocs.io/en/latest/listeners/related-models.html> in crud, that Index and View behave the same, causing less confusion.

If a developer uses crud-view, and prefers ViewListener::beforeFind() <https://github.com/FriendsOfCake/crud-view/blob/master/src/Listener/ViewListener.php#L29> from that, then it is a simple matter of **not** running:
`$this->Crud->listener('relatedModels')->relatedModels([‘Articles’], 'view');`
and thus avoiding the collision, or is that an incorrect understanding?

Since this is an addition and not a change, I don’t see how it can cause problems with existing apps, unless they have run:
$this->Crud->listener('relatedModels')->relatedModels(['Areas'], 'view');
while that part was defunct?

ADmad:
> it is a streamlining of relatedModels in crud, that Index and View behave the same, causing less confusion.

I agree. I am not opposed to this change but we need to consider the possible implication on existing apps.

> I don’t see how it can cause problems with existing apps, unless they have run:
$this->Crud->listener('relatedModels')->relatedModels(['Areas'], 'view');

Well that's not the only way to attach listeners. One can also use `'listeners'` key in `CrudComponent` option and those listeners will be loaded for all actions. So if someone has `Crud.RelatedModels` before `CrudView.View` in the listeners (which is very likely), after this change the contains will be done by the former and the latter will bail out since it checks whether contains are already set.

So we need to think of a way to add this feature without potentially affecting CrudView users.

src/Listener/RelatedModelsListener.php
```diff
@@ -51,6 +52,27 @@ public function beforePaginate(Event $event)
...
 
     /**
+     * Automatically parse and contain related table classes
+     *
+     * @param \Cake\Event\Event $event Before find event
+     * @return void
+     */
+    public function beforeFind(Event $event)
+    {
+        $contained = $event->subject()->query->contain();
```
ADmad: Instead of this code duplication you can just call `$this->beforePaginate()`

ADmad: Or make a put the common code in a protected function `_eventHandler()` and call that from both the callback methods.

eaxexe: Fantastic idea! Should make the code simpler and clearer, I have implemented it and am pushing it in a bit.

ayman-alkom:
+1

jippi:
Stalled for a year :)

PR #449, opened 2016-09-29
CakePHP 3.4: Replace call_user_func* with modern syntax

inoas:
Ref: https://github.com/cakephp/cakephp/wiki/3.4-Roadmap#php-56-minimum
Ref: http://php.net/manual/en/functions.arguments.php#functions.variable-arg-list

### DIFF
--- a/src/Action/AddAction.php
+++ b/src/Action/AddAction.php
@@ -118,8 +118,7 @@ class AddAction extends BaseAction
 
         $this->_trigger('beforeSave', $subject);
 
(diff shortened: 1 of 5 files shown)


ADmad:
Crud supports CakePHP 3.0+ so variadics can't be used and there's no point bumping the version requirement just to be able to use variadics.

ADmad:
My bad, Crud 4.3 support 3.2+ but my above comment still applies.

inoas:
I ll leave it around in my repo, it will sure at some point jump to 3.4.

PR #451, opened 2016-10-05
Fixed bulk action example code

drmonkeyninja:
The example code shows the `_execute()` method rather than the `bulk()` one (I assume this is incorrect based on the plugin's code and the paragraph before the example.


jippi:
Thanks

PR #457, opened 2016-10-28
Fix non-loaded listeners being detected as loaded

bravo-kernel:
This change prevents false positives during Listener detection and thus allows users to now easily switch between listeners (even when config settings for a non-enabled Listener are present in the user application).

Before:
- in my app switch from JsonApiListener to ApiListener
- config setting still present in my app `$this->Crud->config(['listeners.jsonApi.jsonOptions' => [128]]);`
- caused cake error CrudComponent L657, undefined index 'className'

### DIFF
--- a/src/Controller/Component/CrudComponent.php
+++ b/src/Controller/Component/CrudComponent.php
@@ -188,6 +188,7 @@ class CrudComponent extends Component
         $this->_controller->dispatchComponents['Crud'] = true;
 
         $this->_loadListeners();
+
         $this->trigger('beforeFilter');
     }
 
@@ -636,7 +637,7 @@ class CrudComponent extends Component
      * Load a single event class attached to Crud.
      *
      * @param string $name Name
-     * @return \Crud\Listener\BaseListener
+     * @return mixed void|\Crud\Listener\BaseListener
      * @throws \Crud\Error\Exception\ListenerNotConfiguredException
      * @throws \Crud\Error\Exception\MissingListenerException
      */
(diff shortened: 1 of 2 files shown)


bravo-kernel:
Test added

bravo-kernel:
A review/merge would be highly appreciated, switching listeners is killing me right now finishing tests on #445. TIA

bravo-kernel:
ping

src/Controller/Component/CrudComponent.php
```diff
@@ -649,7 +650,12 @@ protected function _loadListener($name)
                 throw new ListenerNotConfiguredException(sprintf('Listener "%s" is not configured', $name));
             }
 
+            if (!isset($config['className'])) {
```
jippi: how does this even happen?

jippi: i would expect us to defer a class name like cake does based on the $name if no className is set

josegonzalez: Seems weird that we'd load the listener if it's not enabled but has configuration. I'm in favor of merging this, as just setting a configuration array doesn't necessarily mean we're loading the thing.

Do we not have some sort of `enabled` key for listeners? I think we do that for actions.

ADmad:
@bravo-kernel Can this be closed?

bravo-kernel:
As always, I will follow you guys so feel free to close (even though I still feel this is very odd default behavior).

PR #458, opened 2016-11-12
Make delete method used configurable

jadb:
Comes handy when using https://github.com/UseMuffin/trash for example.

```
$this->Crud->action()->config('deleteMethod', 'trash');
```

### DIFF
--- a/src/Action/DeleteAction.php
+++ b/src/Action/DeleteAction.php
@@ -30,6 +30,7 @@ class DeleteAction extends BaseAction
         'enabled' => true,
         'scope' => 'entity',
         'findMethod' => 'all',
+        'deleteMethod' => 'delete',
         'messages' => [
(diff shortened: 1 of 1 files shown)


src/Action/DeleteAction.php
```diff
@@ -66,8 +67,9 @@ protected function _post($id = null)
             return $this->_stopped($subject);
         }
 
-        if ($this->_table()->delete($entity)) {
-            $this->_success($subject);
+        $method = $this->config('deleteMethod');
+        if ($this->_table()->$method($entity)) {
+             $this->_success($subject);
```
josegonzalez: extra whitespace here.

jippi:
LGTM

josegonzalez:
Forgot I had commit bit :P

PR #460, opened 2016-11-14
Search listener skip if behavior not loaded on purpose.

dereuromark:
The main controller has
```php
			'listeners' => [
				...
				'Crud.Search',
```
because 90% of the controllers need it.
But if one Table does not have the Search behavior loaded, this means here in this controller we do not need/want the search functionality. Currently it fails with "Missing Search.Search behavior" error.

The exception should only be thrown if the behavior is loaded and something is missing then.

### DIFF
--- a/src/Listener/SearchListener.php
+++ b/src/Listener/SearchListener.php
@@ -53,7 +53,7 @@ class SearchListener extends BaseListener
         }
 
         $table = $this->_table();
-        if (!$table->behaviors()->loaded('Search')) {
+        if (!$table->behaviors()->has('Search')) {
             return;
         }
 
--- a/src/Listener/SearchListener.php
+++ b/src/Listener/SearchListener.php
@@ -53,6 +53,10 @@ class SearchListener extends BaseListener
         }
 
         $table = $this->_table();
+        if (!$table->behaviors()->loaded('Search')) {
+            return;
+        }
+
         if (!$table->behaviors()->hasMethod('filterParams')) {
             throw new RuntimeException(sprintf(
                 'Missing Search.Search behavior on %s',


jippi:
this can lead to odd behavior like `why isn't my search working, there is no errors"  (1h later) oh i forgot to load it in my Table` rather than failing loud and fast.

josegonzalez:
@jippi should we at least log a warning then?

bravo-kernel:
How about making the Exception a bit more explicit. Sth like "Listener configuration detected while listener is not loaded." ?

bravo-kernel:
Apologies, wrong PR. ☕️

dereuromark:
How would one skip then for a single controller?
Couldn't we fail explicitly if no search configuratiin is Provided?

josegonzalez:
@dereuromark [This](https://github.com/josegonzalez/app/blob/master/src/Controller/AppController.php#L89) is what I do in my application skeleton.

dereuromark:
Hmm i don't think we need to protect the dev here that much and that this warrants such Overhead.
But if you want to close the pr, ok

ADmad:
> How would one skip then for a single controller?

You can use CrudComponent::removeListener().

Conversely if only few of your controllers need searching then you can use addListener() instead of configuring it appcontroller itself.

jippi:
Or introduce a SearchAppController with just inject the search stuff - then opt-in is very clear and SearchAppController can just extend AppController :)

agarzon:
Sorry for revive this old thread, but following all from http://crud.readthedocs.io/en/latest/listeners/search.html

No matter what approach I try, always get: 
```
Missing Search.Search behavior on App\Model\Table\ManagersTable 
```

no cache is being used.
behavior is attached in model
```php
$this->addBehavior('Search.Search');
```
https://github.com/FriendsOfCake/search plugin is installed and loaded in bootstrap.php

Any help is welcomed.

PR #461, opened 2016-11-14
Fix doc blocks and return value.

dereuromark:
Some things are not yet fixed:

- "mixed" is used a lot where it could be more concrete probably (string|null etc)

But this already cleans it up quite a bit. Less red/yellow in IDE (phpstorm) as well.

### DIFF
--- a/src/Action/Bulk/BaseAction.php
+++ b/src/Action/Bulk/BaseAction.php
@@ -130,7 +130,6 @@ abstract class BaseAction extends CrudBaseAction
             return $config;
         }
 
-        $ids = $ids;
         $primaryKey = $this->_table()->primaryKey();
         $config['conditions'] = [];
         $config['conditions'][] = function ($exp) use ($primaryKey, $ids) {
@@ -185,8 +184,8 @@ abstract class BaseAction extends CrudBaseAction
     /**
      * Handle a bulk event
      *
-     * @param \Cake\ORM\Query $query The query to act upon
+     * @param \Cake\ORM\Query|null $query The query to act upon
      * @return bool
      */
-    abstract protected function _bulk(Query $query);
+    abstract protected function _bulk(Query $query = null);
 }
--- a/src/Controller/Component/CrudComponent.php
+++ b/src/Controller/Component/CrudComponent.php
@@ -249,7 +249,7 @@ class CrudComponent extends Component
     /**
      * Get a CrudAction object by action name.
      *
-     * @param string $name The controller action name.
+     * @param string|null $name The controller action name.
      * @return \Crud\Action\BaseAction
      */
     public function action($name = null)
@@ -295,7 +295,7 @@ class CrudComponent extends Component
      * To map multiple action views in one go pass an array as first argument and no second argument.
      *
      * @param string|array $action Action or array of actions
-     * @param string $view View name
+     * @param string|null $view View name
      * @return void
      */
     public function view($action, $view = null)
@@ -317,7 +317,7 @@ class CrudComponent extends Component
      * To map multiple action viewVars in one go pass an array as first argument and no second argument.
      *
      * @param string|array $action Action or array of actions.
-     * @param string $viewVar View var name.
+     * @param string|null $viewVar View var name.
      * @return void
      */
     public function viewVar($action, $viewVar = null)
@@ -339,7 +339,7 @@ class CrudComponent extends Component
      * To map multiple findMethods in one go pass an array as first argument and no second argument.
      *
      * @param string|array $action Action or array of actions.
-     * @param string $method Find method name
+     * @param string|null $method Find method name
      * @return void
      */
     public function findMethod($action, $method = null)
@@ -379,7 +379,7 @@ class CrudComponent extends Component
     /**
      * Check if a CRUD action has been mapped (whether it will be handled by CRUD component)
      *
-     * @param string $action If null, use the current action.
+     * @param string|null $action If null, use the current action.
      * @return bool
      */
     public function isActionMapped($action = null)
@@ -487,7 +487,7 @@ class CrudComponent extends Component
      * object.
      *
      * @param string $eventName Event name
-     * @param \Crud\Event\Subject $data Event data
+     * @param \Crud\Event\Subject|null $data Event data
      * @throws Exception if any event listener return a CakeResponse object.
      * @return \Cake\Event\Event
      */
@@ -547,7 +547,7 @@ class CrudComponent extends Component
                 $this->config(sprintf('%s.%s', $type, $realName), $config);
             }
 
-            return;
+            return null;
         }
 
         return $this->config(sprintf('%s.%s', $type, $name));
--- a/src/Listener/RelatedModelsListener.php
+++ b/src/Listener/RelatedModelsListener.php
@@ -70,8 +70,8 @@ class RelatedModelsListener extends BaseListener
      * Find and publish all related models to the view
      * for an action
      *
-     * @param NULL|string $action If NULL the current action will be used
-     * @param NULL|Entity $entity The optional entity for which we we trying to find related
+     * @param null|string $action If NULL the current action will be used
+     * @param null|\Cake\ORM\Entity $entity The optional entity for which we we trying to find related
      * @return void
      */
     public function publishRelatedModels($action = null, $entity = null)
@@ -87,7 +87,7 @@ class RelatedModelsListener extends BaseListener
         foreach ($models as $name => $association) {
             list(, $associationName) = pluginSplit($association->name());
             $viewVar = Inflector::variable($associationName);
-            if (array_key_exists($viewVar, $controller->viewVars)) {
+            if (array_key_exists($viewVar, (array)$controller->viewVars)) {
                 continue;
             }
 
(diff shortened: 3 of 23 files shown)


josegonzalez:
Looks good to me so far.

src/Action/Bulk/BaseAction.php
```diff
@@ -185,8 +184,8 @@ protected function _stopped(Subject $subject)
     /**
      * Handle a bulk event
      *
-     * @param \Cake\ORM\Query $query The query to act upon
+     * @param \Cake\ORM\Query|null $query The query to act upon
      * @return bool
      */
-    abstract protected function _bulk(Query $query);
+    abstract protected function _bulk(Query $query = null);
```
jippi: why did this end up as optional argument?

src/Controller/Component/CrudComponent.php
```diff
@@ -547,7 +547,7 @@ public function defaults($type, $name, $config = null)
                 $this->config(sprintf('%s.%s', $type, $realName), $config);
             }
 
-            return;
+            return null;
```
jippi: `return` is the same as `return null` don't see any benefit to this

src/Listener/RelatedModelsListener.php
```diff
@@ -87,7 +87,7 @@ public function publishRelatedModels($action = null, $entity = null)
         foreach ($models as $name => $association) {
             list(, $associationName) = pluginSplit($association->name());
             $viewVar = Inflector::variable($associationName);
-            if (array_key_exists($viewVar, $controller->viewVars)) {
+            if (array_key_exists($viewVar, (array)$controller->viewVars)) {
```
jippi: why do you need to cast it?

dereuromark: The idea warns about the object. Explicitly Casting makes this here clear.

src/Action/Bulk/BaseAction.php
dereuromark: All extending classes have this optional. Why not the Base class?
Looks like Part of the conctract.

src/Controller/Component/CrudComponent.php
jippi: it's the same... examine this

```
-> php -a
Interactive shell

php > function null_a() { return null; }
php > function null_b() { return; }
php >
php > var_dump(null_a() === null_b());
bool(true)
php >
```

src/Listener/RelatedModelsListener.php
dereuromark: See for yourself in an IDE, viewVars can be arrray or \ArrayObject. Thus the recommended cast.

davidyell: Although I understand, I do worry that guiding the coding style by using a specific third party application will lead to people who do not use that IDE being confused. Shouldn't we instead focus on using `cakephp/cakephp-codesniffer` standard.

dereuromark: This is not a specific third party application, this is a generic thing just pointed out by a tool.
The issue is still valid either way.

src/Controller/Component/CrudComponent.php
dereuromark: It makes it the same quality the core code is. And yes, doc block should also be correct, which it is, right?

src/Listener/RelatedModelsListener.php
jippi:
```
-> php demo.php
object(ArrayObject)#1 (1) {
  ["storage":"ArrayObject":private]=>
  array(1) {
    ["example"]=>
    string(4) "true"
  }
}
bool(true)
bool(false)

cw @ cw.local [ ~ ]
-> cat demo.php
<?php
$a = new ArrayObject();
$a['example'] = 'true';

var_dump($a);
var_dump(array_key_exists('example', $a));
var_dump(array_key_exists('bad_example', $a));
```

dereuromark: "even thought the internal cast may work it is better to be explicit"

jippi: that doesn't even make sense - if the internal cast works, then there is zero reason to be explicit :) its wildly supported all all current php versions works fine https://3v4l.org/gP5qV

dereuromark: IDEs report it - for a reason.
But I am fine reverting this for the sake of compromise.. :/

src/Controller/Component/CrudComponent.php
jippi: since https://wiki.php.net/rfc/void_return_type already made a decission on how this should work, lets just follow that :)

davidyell: This is a plugin, it should not be bound by what the core does.

src/Listener/RelatedModelsListener.php
jippi: yeah, the IDE couldn't be wrong? ;)

src/Controller/Component/CrudComponent.php
dereuromark: It is not bound, but if someone does the effort of doing it right it should not be "shut down" just because, but welcomed IMO :/

src/Listener/RelatedModelsListener.php
dereuromark: They show the symptom - they in this case could be less strict, but why bother?
Easy fix and done.

src/Controller/Component/CrudComponent.php
jippi: Your contributions is always welcome, and the majority is accepted as-is. I or others don't always have to agree with all of the changes of a submitted PR. Open Source is not a democracy, far from it. I'm not dictating what is going on in any FoC project, not even my "own" original projects, but I do voice my concerns when i have them. There has been plenty of PRs that i disliked that have gotten merged into FoC projects, and there will continue to be. But I'm against discussions on semantics where there is an IDE or "upstream" conceptual idea on what's corrects or not. Even if it's the cake core. CRUD is a pretty mature project, and tons of people depend on it, and I'm not in favor of breaking apps or preventing people a zero-change upgrade path, just to fix simple things in the "pure" way over the "hack" way when it can be done. e.g. setter/getter vs magic methods. 

a simple change to the Search plugin caused me to spend more than 1 day fixing all our backend apps since a "semantic" change broke the way we used the plugin. I'm sure plenty of people have had that kind of experience.

dereuromark: You just said it twice that this change does not change anything (thus cannot break anyones app).
I hear you regarding the functional changes, but in this case even you must admit that your concerns are not justified here.

src/Action/Bulk/BaseAction.php
ADmad: Perhaps the extending classes should be fixed instead?  Calling this method without a query instance would make no sense.

src/Controller/Component/CrudComponent.php
ADmad: > But I'm against discussions on semantics where there is an IDE or "upstream" conceptual idea on what's corrects or not.

:+1:  I too don't care for unnecessary code churn unless the changes brings actual value.

src/Action/Bulk/BaseAction.php
dereuromark: That would probably be a BC break, but yeah makes sense.

src/Listener/RelatedModelsListener.php
davidyell: > It is not bound, but if someone does the effort of doing it right it should not be "shut down" just because, but welcomed IMO :/

You're assuming the Core is 'doing it right', you're assuming that you are being 'shut down', and you're assuming your contribution is not welcome. 😉

dereuromark: All I am saying since someone did the work, and it clearly has no argument against it, it is not worth the negative discussion.
Merge, move on.

jippi: its unnecessary and looks bloated compared to the original code, there is more to read and understand, even if the code was the same, i would still prefer the human-readable optimized version over the explicit cast :)

dereuromark: What if (which it does) IDEs show it as yellow/warning?
Then you would wonder if that code actually works or is correct? Everyone would actually focus on that, which doesn't even have to be in the first place.

Right now it is all good and clearly communicated that this is an array afterwards.

src/Controller/Component/CrudComponent.php
inoas: Even if it is not a democracy: ***If*** it doesn't break anything I am all for explicitly returning null as returning an object/data OR void is strange. Used as a setter in future it would be best to just return $this.

However is there is only a slight chance of breaking existing user-land code I am against it.

src/Listener/RelatedModelsListener.php
jippi: i personally don't care about an IDE being wrong, sounds like something that should be fixed upstream :)

src/Controller/Component/CrudComponent.php
dereuromark: There is **not** a slightest change of breaking anything :)

dereuromark:
@josegonzalez @ all others 
Can we merge since all the discussions are around non-issue topics now?

Please also dont forget that someone should open a ticket for removing the nullable default for the next major.

dereuromark:
Guys, can we please continue before this branch gets stale or in conflict state?

jippi:
Sure, please remove the `(array)` and we can get it merged right in :)

dereuromark:
All right. Fair enough.

dereuromark:
Done.

PR #467, opened 2016-12-31
Use CakePHP Router to generate JSONAPI links

bravo-kernel:
This replaces the seriously flawed link generating logic (previously breaking a lot of things) with CakePHP Router generated links. This PR solves most of the concerns I had before. Now:

- respects prefixes
- respects dasherized routes, etc
- respects CakePHP `base` configuration option

Added an option to switch between absolute and relative links as a new year's bonus.


bravo-kernel:
@jippi ping

PR #470, opened 2017-01-11
Remove unnecessary TableRegistry references from JsonApiListener

Marlinc:


ADmad:
Tests are failing

jippi:
I **think** the behavior is strictly not the same as before

`is_a` "`checks if the object is of this class or has this class as one of its parents´" where the new strict equal will not look at the class hierarchy, which mean if someone extends the build-in association types, those will not be detected correctly.  

Using `instanceof ` or an interface would be more robust

ceeram:
Does crud jsonapi listener need to support custom assocation classes? I think the decoupling from ORM is a good thing here personally. I am biased here though as we will use the listener with a non ORM datasource

lorenzo:
The changes look good to me, but tests should pass

bravo-kernel:
Making the listener produce JSON API for all sorts of datasources fits exactly into the megalomanic vision I had when I started work on this, just didn't have a clue where to start so all 👍 ❤️

Marlinc:
I'm working on getting the tests to pass.

src/View/JsonApiView.php
```diff
@@ -213,8 +213,27 @@ protected function _entitiesToNeoMerxSchema(array $entities)
                 throw new CrudException('JsonApiListener cannot not find Entity class ' . $entityName);
             }
 
+            $entityShortName = $entityName;
+            if (strpos($entityName, '\\') !== false) {
```
Marlinc: The entity names that come from the `entiyClass` method are actual full class names. This block of code is necessary to turn those full class names back into the plugin split format which is required by `App::className`.

Marlinc:
The tests are passing now :smile:

lorenzo:
Great job!

bravo-kernel:
@Marlinc I just checked the results in my test API and am seeing an undesired side-effect. The `type` node is now producing full path where it would previously display just `countries`. 

```json
{
  "data": [
    {
      "type": "app\\model\\entity\\countries",
      "id": "1",
      "attributes": {
        "code": "NL",
        "name": "The Netherlands"
    }
  ]
}
```

PR #471, opened 2017-01-11
Make JsonApiListener use schemas as described in the docs

bravo-kernel:
This PR adds the "custom dynamic schema" functionality as described in the docs.

### DIFF
--- a/src/View/JsonApiView.php
+++ b/src/View/JsonApiView.php
@@ -193,8 +193,10 @@ class JsonApiView extends View
     }
 
     /**
-     * Maps entity names to corresponding schema files if it exists, otherwise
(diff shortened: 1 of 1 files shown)


josegonzalez:
Can we specify a schema from a plugin?

bravo-kernel:
Not yet even though I have thought about that. Should be relatively easy but for now making it feature perfect is just too much for me, I'm simply trying to get to the first release here. Ofcourse I am hoping this will get some traction so we can all benefit.

josegonzalez:
No worries, just checking :)

bravo-kernel:
No worries here, just acknowledging that there are other (much) more versed in the plugin area.

PR #472, opened 2017-01-11
Use Cake Router to generate JSON API pagination links

bravo-kernel:
This is the last PR on my list and fixes the pagination links by generating them using CakePHP Router instead of lame hackery (was forgotten in #467).

### DIFF
--- a/src/Listener/ApiPaginationListener.php
+++ b/src/Listener/ApiPaginationListener.php
@@ -87,19 +87,25 @@ class ApiPaginationListener extends BaseListener
      */
     protected function _getJsonApiPaginationResponse(array $pagination)
     {
-        $self = Router::url([
+        $routerMethod = 'normalize'; // produce relative links
+
+        if ($this->_controller()->Crud->config('listeners.jsonApi.absoluteLinks') === true) {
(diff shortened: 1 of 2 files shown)


jippi:
lgtm

src/Listener/ApiPaginationListener.php
```diff
@@ -87,27 +87,33 @@ public function beforeRender(Event $event)
      */
     protected function _getJsonApiPaginationResponse(array $pagination)
     {
-        $self = Router::url([
+        if ($this->_controller()->Crud->config('listeners.jsonApi.absoluteLinks') === true) {
+            $routerMethod = 'url';
+        } else {
```
davidyell: The variables default state could be assigned regardless, and only if the condition matches would it be overwritten, which would remove the `else`. Just a thought 😄

bravo-kernel:
That's what I had originally, reverting 😏

Sent from my Mobile Device

bravo-kernel:
Changed as suggested, rebased, feel free to merge ;)

bravo-kernel:
Thanks!

PR #473, opened 2017-01-14
Changes jsonapi configuration option 'meta' to array type

bravo-kernel:
Changes the type of the `meta` configuration option from **mixed** to **array** as commented by @jippi.


bravo-kernel:
If this could get merged I would be much obliged as it is the basis for the next docs update. TIA

bravo-kernel:
Merci

PR #477, opened 2017-01-20
Remove ORM dependencies and used repository based routing

Marlinc:


Marlinc:
The tests should run but I have to test this a bit more

bravo-kernel:
(Why) are we adding the `.json` extension?

Marlinc:
This was because I was testing in the browser. What are your thoughts on adding the extension based on whether the current URL is being requested with a extension. If it is then we include it in the links as well and if it isn't then we don't?

Marlinc:
Should I create a another PR for that specific part @bravo-kernel ?

ADmad:
codecov show almost 1% drop in coverage. Should that be addressed?

bravo-kernel:
@Marlinc even though I sympathize with the (at initial glance) simpler testing we should IMO **not** support the `.json` extension. The specification explicitely says the Accept header **must** be sent and that is what the Listener enforces atm. May I suggest using Postman for testing, it's the go-to tool for API testing. 

@admad that's why I went for 100% coverage, easy to spot the drops ;)

ADmad:
Fractional drops in coverage when refactoring are common but close to 1% means significant amount of untested stuff :)

bravo-kernel:
Call it an obsession but I only check changed files if they drop below 100% it's time for more tests. 

Wish I could test that `class_exists` in JsonApiListener so that would go to 100% as well but that's for another day, still feels like a 💉 in my eye

ADmad:
@bravo-kernel I'll let you and @Marlinc decide how much coverage is enough :)

bravo-kernel:
@marlinc: one more thing about the extension. We can now use both Api and JsonApi listener at the same time. Api could be triggered by the extension, JsonApi by the Listener.

bravo-kernel:
@Marlinc any indiction when/if you will be able to strip the `.json` extension out of this PR? If you need help or are lacking time I can give it a try so we can keep things moving forward. TIA

Marlinc:
I'll remove it today!

bravo-kernel:
Hero OTD @Marlinc, this totally solves a prefix route issue I was having testing my API blog series. 💯

bravo-kernel:
Could this be merged? Thanks

lorenzo:
Thanks!

PR #483, opened 2017-02-10
Adds propagation stop to add action

bravo-kernel:
This PR adds the possibility to stop propagation for the `add` action which is useful when one requires POST data to be handled by Crud but does not want to create an accompanying database record. 

This is a common requirement for JSON API situations where we MUST use CRUD for everything, even when we do not want/need to create database records. An example:

- Crud JsonApiListener handles JWT authentication
- frontend must POST authentication data to the API
- there is no need for storing the JWT token (and we thus stop propagation)
- we only present the jwt token as JSON API to the frontend for further handling

### DIFF
--- a/src/Action/AddAction.php
+++ b/src/Action/AddAction.php
@@ -116,7 +116,10 @@ class AddAction extends BaseAction
             'saveOptions' => $this->saveOptions()
         ]);
 
-        $this->_trigger('beforeSave', $subject);
+        $event = $this->_trigger('beforeSave', $subject);
+        if ($event->isStopped()) {
+            return $this->_stopped($subject);
+        }
 
         $saveCallback = [$this->_table(), $subject->saveMethod];
         if (call_user_func($saveCallback, $subject->entity, $subject->saveOptions)) {
@@ -165,5 +168,27 @@ class AddAction extends BaseAction
         $this->_trigger('afterSave', $subject);
         $this->setFlash('error', $subject);
         $this->_trigger('beforeRender', $subject);
+    }
+
+    /**
+     * Stopped callback
+     *
+     * @param \Crud\Event\Subject $subject Event subject
+     * @return \Cake\Network\Response
+     */
+    protected function _stopped(Subject $subject)
+    {
+        if (!isset($subject->success)) {
+            $subject->success = false;
+        }
+
+        if ($subject->success) {
+            return $this->_success($subject);
+        }
+
+        $subject->set(['success' => false]);
+        $this->setFlash('error', $subject);
+
+        return $this->_redirect($subject, ['action' => 'index']);
     }
 }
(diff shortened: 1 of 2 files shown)


src/Action/AddAction.php
```diff
@@ -166,4 +169,18 @@ protected function _error(Subject $subject)
...
+
+    /**
+     * Stopped callback
+     *
+     * @param \Crud\Event\Subject $subject Event subject
+     * @return \Cake\Network\Response
+     */
+    protected function _stopped(Subject $subject)
+    {
+        $subject->set(['success' => false]);
```
jippi: wonder if success should default false, but $subject->success could be used to modify it - stopping it could also mean success in some cases?

bravo-kernel: You have a very good point there mr. @jippi and yes, success would definitely be a most valid outcome. Would the following come close? My brain is a bit fried atm.

```php
        $success = false;
        if (!empty($subject->success)) {
            $success = $subject->success;
        }

        $subject->set(['success' => $success]);
```

josegonzalez: Why not an `isset` check instead of `!empty`? Might work better.

bravo-kernel: Thank you mr. @josegonzalez, commit to follow is changed to:

```php
        if (!isset($subject->success)) {
            $subject->success = false;
        }

        $subject->set(['success' => $subject->success]);
```

bravo-kernel:
- success can now be modified using `subject->success`
- added a test for it
- rebased

Let me know if there is anything else.

bravo-kernel:
I'm having problems displaying my manually added record making me wonder if the redirect to `index` is the right approach here. I will investigate further once #477 is merged.

bravo-kernel:
OK Friends, after a lot of testing I think I got this figured out but would appreciate some feedback before updating the PR as I am unsure if this goes against any unknown (underlying) logic. FWIW this:

- allows representing a manually set entity (using both stop and success/true)
- behaves as expected using ApiListener (either empty success or 422, no redirect)
- redirect for non-API

```php
    protected function _stopped(Subject $subject)
    {
        if (!isset($subject->success)) {
            $subject->success = false;
        }

        if ($subject->success === true) {
            return $this->_success($subject);
        }

        $subject->set(['success' => false]);
        $this->setFlash('error', $subject);

        return $this->_redirect($subject, ['action' => 'index']);
    }
```

jippi:
don't think you need the `=== true` part, anything true'ish should be fine - otherwise i think it should be okay?

bravo-kernel:
Thanks, I will update the PR and tests.

bravo-kernel:
Changed and rebased as suggested.

bravo-kernel:
If everybody's in agreement with the suggested change... could it please be merged?

bravo-kernel:
@jippi ping

jippi:
LGTM

bravo-kernel:
Thanks!

PR #493, opened 2017-03-13
Fixed incorrect class inheritance

davidyell:
Seems this example was implementing the Cake Event Interface, whereas elsewhere in the docs it says to extend the BaseListener, which is an abstract class with the `implementedEvents` method already in there.

### DIFF
--- a/docs/listeners/custom.rst
+++ b/docs/listeners/custom.rst
@@ -41,6 +41,7 @@ Creating your own listener class is very similar to using a controller as a list
   namespace App\Lib\Listeners;
 
   use Cake\Event\Event;
+  use Crud\Listener\BaseListener;
 
   class MyListener extends BaseListener
   {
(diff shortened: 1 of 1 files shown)


docs/listeners/custom.rst
```diff
@@ -40,10 +40,9 @@ Creating your own listener class is very similar to using a controller as a list
   <?php
   namespace App\Lib\Listeners;
 
-  use Cake\Event\EventListenerInterface;
   use Cake\Event\Event;
 
-  class MyListener implements EventListenerInterface
+  class MyListener extends BaseListener
```
josegonzalez: Missing the `use` statement?

davidyell: Ah yes, for some reason I saw it was in the same folder, so thought it wouldn't need one. When in fact, it'll not be in the same folder in userland code.

Good spot, thanks!

PR #495, opened 2017-03-16
Adds JsonApiListener disclaimer

bravo-kernel:
Give users fair notice of missing features.

### DIFF
--- a/docs/listeners/jsonapi.rst
+++ b/docs/listeners/jsonapi.rst
@@ -6,10 +6,14 @@ Listener for building a Crud API following the
 
 Introduction
 ------------
-This listener brings you a full implementation of the JSON API specification version
+This listener brings you an implementation of the JSON API specification version
 v1.0 with support for data fetching, data posting, (validation) errors and a ton of
 configurable options to manipulate the generated json allowing you to benefit of instant
 compatibility with JSON API supporting tools and frameworks like Ember Data.
+
+Please note that some parts of the JSON API specification have not been implemented yet.
+Feel free to submit a PR for missing functionality and help work towards a full-featured
+implementation of the specification, the effort should be minimal.
 
 Requirements
 ------------


Spriz:
Is it already known what's the missing parts? That would make it easier to implement the "rest"! :)

josegonzalez:
@Spriz its not. If you have a list of those, feel free to create issues.

bravo-kernel:
I will come up with a list. Nice pun mr. @Spriz, you json of a...

bravo-kernel:
Going top-to-bottom through the spec I see four things are missing:

- **Top level `links` node:** http://jsonapi.org/format/#document-links
- **Sparse fieldsets**: http://jsonapi.org/format/#fetching-sparse-fieldsets)
- **Sorting**: http://jsonapi.org/format/#fetching-sorting
- **Filtering**: http://jsonapi.org/format/#fetching-filtering

There is a fifth one ([Query Parameters](http://jsonapi.org/format/#query-parameters),) which is no specification but covers implementation specific query parameters. I always felt this could be really useful for manipulating the config options now only settable using the listener config (e.g. enabling/disabling the jsonapi-version-node with a query parameter). However, I'm unsure if this would possibly interfere with other query parameter based plugins (like e.g. search).

To end this in a constructive manner... I could pick up the first one as it was once in the listener but probably got deleted somewhere along the way. The other three missing parts would require someone diving into it as they are both not my expertise and I don't need them at this point (my frontend-app works fine as is atm).

Spriz:
😅  Sorry 'bout that!

bravo-kernel:
Haha, no apologie needed. I interpreted it as the most positive, seriously. Even though I later considered you might not even intended it like that 😱  You are completely on my good list mr. @Spriz

bravo-kernel:
Created tickets for remaining missing functionality #521, #522, #523.  Skipped both "filtering" and "query parameters" as they look debatable.

dakota:
Filtering is/could already be supported by using the search listener?
Is sort not already supported too?

bravo-kernel:
That might very well be the case mr. @dakota. I created the tickets to make sure it works (and to prevent any possible future noise). I will test and try today if I find some time. Thanks for the pointers.

bravo-kernel:
Never realized they meant search with `filter` 🙄. I might create a ticket just to make sure we are complete.

PR #502, opened 2017-03-23
Do not use/set ext for jsonapi detector

bravo-kernel:


bravo-kernel:
Copy/paste left-over from ApiListener. Not required as the JSON API spec does not use extensions.

Issue #494, opened 2017-03-14
relatedModels listener does not work for Crud.view method


phishy:
Is there a reason this isn't supported out of the box? I easily added a beforeFind method to RelatedModelsListener and registered the event and it works fine. Just curious before I submit a pull request.

josegonzalez:
None that I can think of.

josegonzalez:
https://github.com/FriendsOfCake/crud/pull/448

phishy:
Yeah, seems about right.

ayman-alkom:
+1

viniciusbig:
So, I saw #448 and #505 and I couldn't understand this.

I'm using Crud plugin without CrudView.

I've loaded the listeners and actions in my AppController to make it global.

My Clients model belongsTo an Account

So, using a regular view controller, I can get the account like this:
`$client = $this->Clients->get($id, [ 'contain' => ['Accounts'] ]);`

but using the default View action from CRUD, I don't have this association.

Even loading the relatedModels in the view action

```
$this->loadComponent('Crud.Crud', [
    'actions' => [
        'view' => [
            'className' => 'Crud.View',
            'relatedModels' => true
        ],
    ],
    'listeners' => [
        'relatedModels',
        'Crud.RelatedModels',
    ]
]);
```

How can I use the view action and still get the associated data?

viniciusbig:
Sorry, why this is closed?
Is the related models on View actions working?
Just asking because I'm still having issues with that

PR #505, opened 2017-03-26
make relatedModels works with Crud.View

ayman-alkom:
refs: #448 , #494 
in most cases it is more necessary in the view than the index

### DIFF
--- a/src/Listener/RelatedModelsListener.php
+++ b/src/Listener/RelatedModelsListener.php
@@ -25,6 +25,7 @@ class RelatedModelsListener extends BaseListener
     {
         return [
             'Crud.beforePaginate' => 'beforePaginate',
+            'Crud.beforeFind' => 'beforeFind',
             'Crud.beforeRender' => 'beforeRender',
         ];
     }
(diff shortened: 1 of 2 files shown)


ADmad:
Opening a new PR with same changes as done in #448 doesn't help. I have already commented there why the PR cannot be merged in it's current form.

ayman-alkom:
Hi @ADmad 
I made a new PR
- because the old one seems stopped and in the same time open, in other words it can be acceptable if it continued the work.
- this issue seem like a wanted feature, it is even expected and I see that some people spent some hours to run the feature then find it not made, I also had that.
- The old PR seems not passing the tests, it may need rebase
I didn't touched the tests, if you are agree with this feature and PR, then we can talk about the tests.

ADmad:
@ayman-alkom That PR is stalled because the issue I raised was not addressed. It would have been better if you had first started a dialogue on how to made changes instead of opening a new PR with exact same modifications.

One way to address the concern regarding crud view is allow the events for which the containments are done by `RelatedModelsListener` to be configurable and have it disabled by default for `beforeFilter`. Then we can just document how to enable it for `beforeFilter` too.

ADmad:
Another bigger concern is that containing associated models for `beforeFind` means related records will be fetched for edit action too. In best case scenario it would cause unnecessary fetching of related records and in worst case scenario cause unwanted/unexpected updating of records.

ayman-alkom:
Hi @ADmad , the issues is clear now, I am agree with you, but I didn't use the CrudView before so I don't know a lot  about it, but that is not a problem.
I will take a look to the CrudView and a deeper look to the event configuration and see how we can address the issues you raised above.
you can close this PR, and open a new issue for the dialogue or keep it open so I will submit here.

PR #514, opened 2017-03-31
Add SerializeTrait to EditAction

alysson-azevedo:
EditAction should use SerializeTrait.
http://crud.readthedocs.io/en/latest/actions/edit.html#serialize


ADmad:
Why?

alysson-azevedo:
Because [AddAction](https://github.com/FriendsOfCake/crud/blob/master/src/Action/AddAction.php#L22) has it and docs says EditAction should have too.

ADmad:
Okay. Please add tests for the change.

PR #517, opened 2017-04-04
JsonApi updates http code 422 and 409 cannot send correct headers.

lpj145:
When validation errors appear, headers cannot send correct, and data is blocked because not have anyone header.

### DIFF
--- a/src/Error/JsonApiExceptionRenderer.php
+++ b/src/Error/JsonApiExceptionRenderer.php
@@ -97,6 +97,10 @@ class JsonApiExceptionRenderer extends \Cake\Error\ExceptionRenderer
         // set data and send response
         $this->controller->response->type('jsonapi');
         $this->controller->response->body($json);
+        $this->controller->response->cors($this->controller->request)
+            ->allowOrigin('*')
+            ->allowCredentials()
+            ->build();
 
         return $this->controller->response;
     }


src/Error/JsonApiExceptionRenderer.php
```diff
@@ -97,6 +97,10 @@ public function validation($exception)
         // set data and send response
         $this->controller->response->type('jsonapi');
         $this->controller->response->body($json);
+        $this->controller->response->cors($this->controller->request)
```
ADmad: I don't think exception render is the correct place to set CORS headers. That should be done for your request URL in middleware or beforeFilter ().

lpj145: Sorry yesterday i have problem with my mind.... :smile: i used plugin cors to do it, but when have error CORS no works correctly.

lpj145: But the ozee31/Cors plugin is overload by crud plugin and return headers have an difference. I think  this crud plugin dont needed cors enabled ? this wold be possible if config on enable plugin, what do you think about this ?

ADmad: It's upto you to set appropriate CORS headers for your app. The plugin won't do it for you.

ADmad:
As I mentioned on slack, the problem most likely lies in the middleware of [ozee31/cors](https://github.com/ozee31/cakephp-cors/blob/master/src/Routing/Middleware/CorsMiddleware.php#L14) you are using. Response instance should be modified *after* calling `$next()`.

bravo-kernel:
Not Crud related, solved in https://github.com/ozee31/cakephp-cors/releases/tag/v1.1.0

PR #518, opened 2017-04-06
Json api Error if the entity class is \Cake\ORM\Entity

dakota:
If the entity class for a repository is the default `\Cake\ORM\Entity` the the JSONAPI parser fails with a cryptic "Invalid schema" error. This causes the json api view class to throw a more verbose and descriptive error in that case to hint to the developer what they should be looking at.


bravo-kernel:
@dakota is this PR still desired? If so I can try to fix once #520 gets merged.

bravo-kernel:
One test failing on PHP 5.5

src/View/JsonApiView.php
```diff
@@ -212,6 +213,10 @@ protected function _entitiesToNeoMerxSchema(array $repositories)
 
             $entityClass = $repository->entityClass();
 
+            if ($entityClass === Entity::class) {
+                throw new CrudException(sprintf('Entity classes must be the generic "%s" class for repository "%s"', $entityClass, $repositoryName));
```
ADmad: ..must *not* be.. ?

ADmad:
Test failing.

bravo-kernel:
Fix in progress

bravo-kernel:
A new release would be nice, I feel a bit hesitant to create one.

PR #519, opened 2017-04-11
Added haders from older controller

lpj145:
When have exception, all headers are destroyed and new Response instance created, with this, new Response set olders headers.


ADmad:
Simply copying headers of response instance of current controller onto response instance of error controller could lead to undesired results.

I already explained on your previous PR that the problem lies in the ozee31/cakephp-cors plugin you are using. So please divert your efforts to getting that fixed instead. I have even opened an issue ozee31/cakephp-cors#8 for your benefit.

bravo-kernel:
Confirmed: not Crud related, solved in https://github.com/ozee31/cakephp-cors/releases/tag/v1.1.0

PR #530, opened 2017-05-05
Jsonapi split

bravo-kernel:
Prepares splitting JSON API to https://github.com/FriendsOfCake/crud-json-api


bravo-kernel:
Just tested the split in my test-app and works as expected after changing the listeners to:

```php
            'listeners' => [
                'CrudJsonApi.JsonApi',
                'CrudJsonApi.Pagination',
                'Crud.ApiQueryLog'
            ]
```

Stripping out the docs next.

jippi:
why? :)

dakota:
There is still a fair amount of work that needs to happen with the json API stuff, this allows it to be developed without affecting the core CRUD release cycle

ADmad:
@jippi The JsonApi related code gets regular updates while rest of the Crud is pretty stable. Having it separate makes maintenance easier and releases for both can be done independently as @dakota said.

bravo-kernel:
@jippi it was requested exactly for the reasons above, I simply follow(ed)

bravo-kernel:
Just to say.... I'm fine with not splitting and just deleting this PR too. Seems I've hitten the point were I feel I can have an opinion so no work lost either way ;)

bravo-kernel:
- the [crud-json-api](https://github.com/FriendsOfCake/crud-json-api) repo is all set up
- docs are moved to https://crud-json-api.readthedocs.io/en/latest/

Everything is good to go as far as I can see, please let me know how to proceed.

ADmad:
I would prefer having jsonapi separate.

bravo-kernel:
If moved it would be nice if the four issues could be moved to the new repo as well. Has anybody ever pulled that off (successfully) before?

ADmad:
I can take care of moving the issues.

ADmad:
@jippi Any objections to merging this?

ADmad:
Everything is setup on the new repo including docs for it on readthedocs.org.

jippi:
no objections :) 

the only downside as i see the split will be that crud activity will look "stale" (since it's so stable and relatively bug free by now), and that can in some cases harm adoption rate.

not sure it would be a bad thing to tie crud and crud-json-api release cycle anyway, for pure-crud users it would just mean no problems :) 

but if you prefer the split, i have no objections at all :)

ADmad:
We could come up with some new ideas to keep this repo look busy 😀

ADmad:
@bravo-kernel  i'll take care of moving related issues to new repo later today.

josegonzalez:
@bravo-kernel for reference, this is useful: https://github-issue-mover.appspot.com/

davidyell:
Splitting seems like a good idea to me, I agree with the points raised here. It does make me wonder if the main Crud repo docs, needs a section explaining about the 'optional extras' which can be mixed in to Crud.

ADmad:
A section in docs mentioning crud's ancillary plugins like jsonapi, crud-users, crud-view would be nice. We can also add them to composer's "suggests" section.

bravo-kernel:
Totally agree with making crud plugins more visible by integrating them in the Crud docs, "suggests" are nice but end-users will probably read the docs mostly IMO. I feel #531 could be a good example.

josegonzalez:
This should be updated in all FOC projects: https://github.com/FriendsOfCake/crud/blob/master/docs/conf.py#L25

bravo-kernel:
Ideally with this list so JsonApi is included as well: https://github.com/FriendsOfCake/crud-json-api/blob/master/docs/conf.py#L25

josegonzalez:
Might be worth just exposing a json endpoint on the foc site (github pages) that has a list of all FOC projects, and then curling it in that conf.py. That way we can expand the footer as necessary.

bravo-kernel:
That would be more than awesome 💯

PR #533, opened 2017-05-26
_checkRequestMethods should throw MethodNotAllowedException

### LINKED ISSUE #409: _checkRequestMethods should throw MethodNotAllowedException
CakePHP has an exception MethodNotAllowedException that returns the HTTP error code 405 (Method Not Allowed). It seems to me that **ApiListener::_checkRequestMethods** should use this exception and error code rather than the more generic BadRequestException (with its standard error code of 400).

Spriz:
Closes #409

### DIFF
--- a/src/Listener/ApiListener.php
+++ b/src/Listener/ApiListener.php
@@ -4,7 +4,7 @@ namespace Crud\Listener;
 use Cake\Core\Configure;
 use Cake\Error\ErrorHandler;
 use Cake\Event\Event;
-use Cake\Network\Exception\BadRequestException;
+use Cake\Network\Exception\MethodNotAllowedException;
 use Cake\Network\Request;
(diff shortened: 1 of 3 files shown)


ADmad:
Tests failing.

src/Listener/ApiListener.php
```diff
@@ -149,7 +149,7 @@ protected function _checkRequestMethods()
             }
         }
 
-        throw new BadRequestException('Wrong request method');
+        throw new MethodNotAllowedException('Wrong request method');
```
ADmad: The message string can also be removed and instead just let the default message for this exception be used.

Spriz:
Oh the joy of editing code from Github 👍

Spriz:
there we go @admad :)

PR #534, opened 2017-05-26
Move logic for dispatchComponents to ::initialize to comply with Cake 3.3 AuthComponent

Spriz:
With this change: https://github.com/cakephp/cakephp/pull/6906 one can have the AuthComponent working in `Controller::initialize` rather than `Controller::beforeFilter` - but if one does that with crud, the `dispatchComponents`property is not yet filled, and because of that the code in the ControllerTrait doesn't work. (https://github.com/FriendsOfCake/crud/blob/master/src/Controller/ControllerTrait.php#L96).

This makes the AuthComponent *skip* both authentication and authorization, unless there is boilerplate code as follows for each action (because of: https://github.com/cakephp/cakephp/blob/master/src/Controller/Component/AuthComponent.php#L291)

```
<?php
namespace Controller;

use Cake\Event\Event;

class PostsController extends AppController
{
    public function index()
    {
    }

    public function add()
    {
    }

    public function edit($id)
    {
    }

    public function delete($id)
    {
    }

    public function view($id)
    {
    }
}

```

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -121,6 +121,7 @@ If you've loaded an action in eg. your ``AppController`` - but don't want it inc
 Example of disable a loaded action:
 
 ..code-block:: phpinline
+
   class AppController extends \Cake\Controller\Controller {
 
     public $components = [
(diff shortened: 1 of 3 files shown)


Spriz:
I was unfortunately not able to make a failing test case - but maybe someone else could assist me? :)

Spriz:
Just wanted to point out this has been working always without the "boilerplate" code with just `return $this->Crud->execute()` and seem to have ever since Cake 2.x

PR #541, opened 2017-08-02
Phpstan

ADmad:


src/Action/AddAction.php
```diff
@@ -126,7 +126,7 @@ protected function _post()
             return $this->_success($subject);
         }
 
-        return $this->_error($subject);
+        $this->_error($subject);
```
josegonzalez: Damn this actually breaks a change I have where I redirect back to a different page on error.

Any thoughts on how to do that with this change?

ADmad: We can change return type of `_erro()` to `Reponse|null` and add back this `return`.  That should make phpstan happy.

ADmad:
@josegonzalez I added back the `return` for `_error()` calls.

Issue #547, opened 2017-08-28
PHP 7.2: Rename Object class


inoas:
> "object" (in any case) can no longer be used as a class name.

Source: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2907#issue-241202093
Ref: https://wiki.php.net/rfc/object-typehint

In Crud: https://github.com/FriendsOfCake/crud/blob/master/src/Core/Object.php#L16
PHPParser names its Object class `Object_` - should we do the same?

inoas:
Closing as PR is open and the issue is discussed over there.

PR #548, opened 2017-08-28
Rename class Object to Object_ for PHP 7.2 compatibility

inoas:
Issue is here: https://github.com/FriendsOfCake/crud/issues/547


davidyell:
I would rather just change the name of the class to something like `Base` personally. Adding an underscore seems lazy and unnecessary.

inoas:
@ `Base` vs `Object`:

- People familiar with the internals will look for `class Object` and they will find it, either as a class alias or as `Object_` - migration of forks et al. should be easy.
- `Base` is far to generic. `BaseObject` could work; I still prefer `Object_` as for a release that still maintains a class_alias() to `Object`.

@ new major release and re-working of the internals of the plugin:

- I agree then things could be changed beyond - maybe not having an Object class or renaming it to `BaseObject` and also droppping the deprecation?
- Changing the Plugins internals is beyond the scope of this PR.
- PHP 7.2.0 stable is around the corner and things need to happen rather quickly, else CRUD will break on a stable, maintained, future-forward PHP release. That's the reason I created this PR.

ADmad:
Renaming to `BaseObject`  along with using `class_alias()` for old name sounds reasonable.

inoas:
Two member votes, so I did it.

Keep in mind that I am not very familiar with CRUD internals. The tests say it is okay, but maybe you guys want to give it a try too. Otherwise from my side this is GTM now.

ADmad:
Please also add `7.2` to `.travis.yml` so that tests are run on PHP 7.2

ADmad:
Hmm.. Class renames are handled a bit differently in core. Taking the example of renaming [Cake\Network\Response](https://github.com/cakephp/cakephp/blob/master/src/Network/Response.php) to [Cake\Http\Response](https://github.com/cakephp/cakephp/blob/master/src/Http/Response.php) the `class_alias()` call in added to the new class and old class has a `clas_exists()` call for the new class.

inoas:
This has been my pattern that I replicated from the cakephp/cakephp core:
https://github.com/cakephp/cakephp/blob/master/src/Utility/String.php#L1-L5

inoas:
Is PHP 5.5 still supported in CRUD?

ADmad:
> This has been my pattern that I replicated from the cakephp/cakephp core:
https://github.com/cakephp/cakephp/blob/master/src/Utility/String.php#L1-L5

Okay, if all builds pass then we can keep this way. The good thing is it's highly unlikely someone would be directly using the `Object` class, so this renaming shouldn't cause any problems.  Plugins/apps using Crud would be extending `BaseListener` or `BaseAction` which extend `Object`.

> Is PHP 5.5 still supported in CRUD?

Crud supports Cake 3.2+ so yes it supports PHP 5.5.

ADmad:
Build [failing](https://travis-ci.org/FriendsOfCake/crud/jobs/269990058#L451) on 7.2. You need to update .travis.yml to not try to remove php.ini for 7.2 as it doesn't exist.

ADmad:
Wonder why renaming the class causes this huge drop in code coverage.

inoas:
> Build failing on 7.2. You need to update .travis.yml to not try to remove php.ini for 7.2 as it doesn't exist.

I don't know much about travis, would this work? https://github.com/cakephp/cakephp/blob/3.next/.travis.yml#L37

I can move the class alias around if you think it helps with the coverage...?

ADmad:
Closing in favor of #549. Thank for reporting this.

PR #549, opened 2017-08-30
Rename "Object" to "BaseObject" for PHP 7.2 compatibility.

ADmad:


inoas:
Thanks for the quick response

PR #552, opened 2017-10-09
Adds propagation stop to edit action

johanmeiring:
Based on work that was originally done here: #483.

I've run into a situation where I need to stop a save _before_ it actually happens, based on factors external from Crud, which are checked for in a `beforeSave` event.  This PR makes that possible.


ADmad:
Can you elaborate on your use case which required stopping the save action?

johanmeiring:
@ADmad Sure :-)  Two of the fields in my table form a unique index.  Currently, if a record is updated with data which causes it to violate the uniqueness constraint of an existing record, a database exception is thrown.  With the proposed change, I will be able to check for the existence of such a record first (in a `beforeSave` function), and then stop event propagation if such a record already exists.

jippi:
sounds like a normal validation rule on [`isUnique`](https://book.cakephp.org/3.0/en/orm/validation.html#creating-unique-field-rules)

johanmeiring:
@jippi Good point, I didn't think of that.  I'll try it.  But aside from that, I'm sure this could still be useful to someone for some other reason in future?

johanmeiring:
I think a good reason for adding it anyway is because it adheres to what I've observed to be the CakePHP convention of being able to stop a workflow in its tracks at any point simply by issuing `$event->stopPropagation()`... Based on said observation, I expected that to work in this case, but it didn't.

dakota:
I think this is a good idea, if only to match `AddAction`.

ADmad:
If someone provides a valid use case example I would have no problem with this but I am hesitant to add code just based on the premise that someone *might* need it.

josegonzalez:
I've got a use case! In my weird `calico` cms, I need to fail the `put` when I am unable to validate an entity. Why am I using the `Crud.beforeSave` trigger to validate instead of the regular `Table.beforeSave`? Because I need to rewrite all the data to actually not be in the "wrong" format - to rewrite it so that it doesn't look as if its in the same entity, but in a related entity.

src/Action/EditAction.php
```diff
@@ -118,7 +120,13 @@ protected function _put($id = null)
             $this->saveOptions()
         );
 
-        $this->_trigger('beforeSave', $subject);
+        $event = $this->_trigger('beforeSave', $subject);
+        if ($event->isStopped()) {
+            $subject->set(['success' => false]);
+
+            return $this->_stopped($subject);
```
ADmad: `_stopped()` redirects to index action. Shouldn't a failed save render the form instead with submitted data as it's done when `Model.beforeSave` fails?

josegonzalez: So should `_stopped()` have a second arg that allows disabling the redirect?

ADmad: I guess so.

jippi:

PR #560, opened 2017-12-28
Allow using custom finder with options.

ADmad:

PR #563, opened 2018-01-04
Allow all actions to use custom finder with options.

ADmad:
Follow up of #560.


ADmad:
Earlier I overlooked the fact that `FindMethodTrait::_findRecord()` isn't used actions which work multiple records.

PR #573, opened 2018-01-31
Fix #572 added api key to defaultconfig in view and lookup actions

### LINKED ISSUE #572: ViewAction and LookupAction is missing api index in $_defaultConfig
From version 5.3 the method `Crud/Listener/ApiListener::respond` is changed to call `$response->withStatus` in stead of `$response->statusCode`. Because the actions View and Lookup don't have the key `api`, the status code in the Response is set to null. CakePHP doesn't check if null is a valid status code and add it to the response. This will result in a 502 bad gateway error in Nginx for example.
There is already a discussion going in cakephp/cakephp (https://github.com/cakephp/cakephp/issues/11677), but it would be advisable to make sure the key `api` exists in the ViewAction and LookupAction.
A pull request will be created shortly.

frankvanhest:
See issue https://github.com/FriendsOfCake/crud/issues/572.
This will add the `api` key to the $_defaultConfig in the ViewAction and the LookupAction.

### DIFF
--- a/src/Action/LookupAction.php
+++ b/src/Action/LookupAction.php
@@ -28,7 +28,15 @@ class LookupAction extends BaseAction
     protected $_defaultConfig = [
         'enabled' => true,
         'scope' => 'table',
-        'findMethod' => 'list'
+        'findMethod' => 'list',
+        'api' => [
+            'success' => [
+                'code' => 200
+            ],
+            'error' => [
+                'code' => 400
+            ]
+        ]
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


ADmad:
@frankvanhest PR #574 should fix the issue. Adding this status codes to view and lookup actions isn't necessary since 200 is the default for success and there would be an error only if a query failed and in that cause the exception handler/render would take of returning response with proper code.

bravo-kernel:
🚜: mimic PR for JsonApiListener

frankvanhest:
@ADmad That is even better that changing the default configs. Your way falls back to the default status code within CakePHP.

PR #575, opened 2018-02-03
feat: allow users to override what happens when a `recordNotFound` exception is thrown

josegonzalez:


jippi:

PR #576, opened 2018-02-07
ApiPaginationListener should use setConfig in beforeRender

frankvanhest:
After updating to crud tag 5.3.x the pagination failed due to a faulty replacement of the deprecated method InstanceConfigTrait::config(). 
At Crud\Listener\ApiPaginationListener line 71 the deprecated config method was used as a setter, but it was replaced as a getter.

### DIFF
--- a/src/Listener/ApiPaginationListener.php
+++ b/src/Listener/ApiPaginationListener.php
@@ -68,6 +68,6 @@ class ApiPaginationListener extends BaseListener
         ];
 
         $controller->set('pagination', $paginationResponse);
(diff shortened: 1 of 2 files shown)


ADmad:
Looks good, can you add a test case to avoid regression?

frankvanhest:
Sure, I'll write a test for it. It will be provided tomorrow evening (GMT)

frankvanhest:
@ADmad A test has been added. Also the current tests for the ApiPaginationListener are updated so they work again after this change.

ADmad:
@frankvanhest Thank you, merged and new release done.

PR #579, opened 2018-02-25
Fix compatibility with older version of search plugin.

### LINKED ISSUE #578: Search listener breaks
I noticed in last month update of `search` listener where it used `filterParams()` function it was changed to `$repository->behaviors()->filterParams($this->_request()->getQuery());` instead of calling it from `Table` directly. The problem with this is that the `if()` check where it checks if method exists is not good enough and it goes trough even though it doesnt have it.
pic1
pic2

ADmad:
Closes #578.

@itodor Can you confirm this fixes your problem?


itodor:
@ADmad yes this helps, thx

PR #587, opened 2018-04-15
Cake 36

ADmad:


bravo-kernel:
This makes me wish crud-json-api was still a part of Crud

PR #589, opened 2018-04-23
Lower CakePHP version requiment to 3.5.

ADmad:


ADmad:
Update to .travis.yml coming up for testing against lowest version.

ADmad:
Done

josegonzalez:
Why this change @ADmad ?

ADmad:
So that current minor can work with Cake 3.5 as well as 3.6 without issues/warnings and bumping min. Cake version dependency to 3.6 can be avoided.

PR #595, opened 2018-06-18
Require CakePHP 3.6 since we use 3.6-only features

Spriz:
https://github.com/FriendsOfCake/crud/releases/tag/5.4.2 Includes 3.6 exceptions - so I guess we should also require 3.6 as the minimum.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         }
     ],
     "require": {
-        "cakephp/cakephp": "^3.5"
+        "cakephp/cakephp": "^3.6"
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7.14|^6.0",


Spriz:
Hmm - should it be `"cakephp/cakephp": ">=3.6"` ?

ADmad:
Version bump should not be required. Yes the exception classes are used from `Http` namespaces but the plugin has [aliases.php](https://github.com/FriendsOfCake/crud/blob/master/aliases.php) which alias the new class under the old name for Cake 3.5 compatibility.  And this aliases file is loaded by [composer](https://github.com/FriendsOfCake/crud/blob/master/composer.json#L55).

Spriz:
I see. Didn't notice that file! Full stack trace for what it's worth: https://gist.github.com/Spriz/ee29d5fe1aef52e42a5c75cbac09720e

PR #596, opened 2018-06-18
add alias for 'NotFoundException'

saeideng:
Refs : https://github.com/FriendsOfCake/crud/pull/595


aliases.php
```diff
@@ -13,7 +13,7 @@ class_alias(
         'Cake\Http\Exception\BadRequestException'
     );
     class_alias(
-        'Cake\Network\Exception\NotFoundException'
-        'Cake\Http\Exception\NotFoundException',
+        'Cake\Network\Exception\NotFoundException',
+        'Cake\Http\Exception\NotFoundException'
```
davidyell: Be nice when we get PHP 7.2 as trailing commas will be allowed everywhere.

PR #601, opened 2018-09-13
Add SerializeTrait to View and Edit Actions

raul338:
The code in the docs throw exceptions
```php
$this->Crud->action()->serialize();
// Call to undefined method Crud\Action\ViewAction::serialize()
```
This PR fix that

### DIFF
--- a/src/Action/EditAction.php
+++ b/src/Action/EditAction.php
@@ -6,6 +6,7 @@ use Crud\Event\Subject;
 use Crud\Traits\FindMethodTrait;
 use Crud\Traits\RedirectTrait;
 use Crud\Traits\SaveMethodTrait;
+use Crud\Traits\SerializeTrait;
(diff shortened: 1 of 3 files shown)


ADmad:
Tests needed.

raul338:
Tests added. I add a simple test for all actions

raul338:
Should I do anything else?

ADmad:
Sorry but your tests are pointless. You should be asserting that view var `_serialize` contains the key you set when calling `serialize()` method. And since the action's serialize config is used only by `ApliListener` your tests should be in `ApiListenerTest`.

raul338:
Ok, I'll try on the weekend

raul338:
Should I do something else?

raul338:
Could you release a bugfix release please?

ADmad:
Done

PR #606, opened 2018-11-21
RelatedModelsListener - missing oneToMany relation

wsergent:
Added oneToMany to the RelatedModelsListeners as without it, our hasMany() relations were not showing up when using Crud

### DIFF
--- a/src/Listener/RelatedModelsListener.php
+++ b/src/Listener/RelatedModelsListener.php
@@ -147,7 +147,7 @@ class RelatedModelsListener extends BaseListener
         $settings = $this->relatedModels(null, $action);
 
         if ($settings === true) {
-            return $this->getAssociatedByType(['oneToOne', 'manyToMany', 'manyToOne']);
+            return $this->getAssociatedByType(['oneToOne', 'manyToMany', 'manyToOne', 'oneToMany']);
         }
 
         if (empty($settings)) {
--- a/tests/TestCase/Listener/RelatedModelsListenerTest.php
+++ b/tests/TestCase/Listener/RelatedModelsListenerTest.php
@@ -117,7 +117,7 @@ class RelatedModelListenerTest extends TestCase
         $listener
             ->expects($this->once())
             ->method('getAssociatedByType')
-            ->with(['oneToOne', 'manyToMany', 'manyToOne']);
+            ->with(['oneToOne', 'manyToMany', 'manyToOne', 'oneToMany']);
 
         $result = $listener->models();
     }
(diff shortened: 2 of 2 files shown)


davidyell:
Test cases need to be updated to reflect this change.

ADmad:
Having all has many associated records pulled in by default is not something most (including me) would want. It can add a significant overhead of number of queries and memory required. So I am not in favor of this change.

I am open to having a way to easily configure the association types that should be fetched by default. Currently it's only possible to specify models by names not by association type.

ADmad:
@wsergent Not a biggie but adding `+1` reaction to your own PR isn't really useful :slightly_smiling_face:.

wsergent:
Thanks @ADmad - perhaps I am doing something wrong in the way I am attempting to fetch the RelatedModels?
I couldn't get them to show up in my view() function, even using a contain(), without making this change to the `crud` vendor files.

Here is my view() function from my VenuesController. I can supply the model definitions if it helps, but there's nothing too complex. In fact, VenuesTable hasMany() for all of these relationships except Timeslots which is a belongsToMany().


`
    public function view($id) {
        $this->Crud->on('beforeFind', function(\Cake\Event\Event $event) {
            $event->getSubject()->query->contain([
                'Timeslots',
                'Dishes',
                'Proteins',
                'Sides',
                'Sauces'
            ]);
        });
        return $this->Crud->execute(null, ['id' => $id]);
    }
`

ADmad:
By default the `ViewAction` doesn't even use `RelatedModels` listener. Did you enable it yourself for the action? Are you using just the Crud plugin or CrudView too?

wsergent:
In my AppController I define `$components` as thus:

```
    public $components = [
        'RequestHandler',
        'Crud.Crud' => [
            'actions' => [
                'index' => [
                    'className' => 'Crud.Index',
                    'relatedModels' => true
                ],
                'view' => [
                    'className' => 'Crud.View',
                    'relatedModels' => true
                ],
                'add' => [
                    'className' => 'Crud.Add',
                    'relatedModels' => true
                ],
                'edit' => [
                    'className' => 'Crud.Edit',
                    'relatedModels' => true
                ],
                'Crud.Delete'
            ],
            'listeners' => [
                'Crud.Api',
                'Crud.ApiPagination',
                'Crud.ApiQueryLog',
                'Crud.RelatedModels'
            ]
        ]
    ];
```

Everything else (so far) has worked as expected except for the `ViewAction`, hence why I used the `beforeFind` to contain my missing models.

ADmad:
Okay so you have enabled the listener for `ViewAction`. Now the `RelatedModels` listener only runs for `beforePaginate` and `beforeRender` events and in case of `ViewAction` it would run for `beforeRender`.

For `beforeRender` the `RelatedModels` listener populates view variables doesn't modify the find query for primary record at all, so it's related to the containments you are doing to the query hence I don't see how this changes you have made are relevant.

Also the callback you are setting for `beforeFind` to modify the query seems fine to me, have you ensured that callback function is executed?

jippi:

PR #624, opened 2019-06-25
ApiQueryLogListener eats disk space even when debug=false

asgraf:
ApiQueryLogListener::setupLogging() event should not enable query logging if debugging is not enabled. Please note that enabling query logging causes dumping those logs into queries.log file

### DIFF
--- a/src/Listener/ApiQueryLogListener.php
+++ b/src/Listener/ApiQueryLogListener.php
@@ -49,6 +49,10 @@ class ApiQueryLogListener extends BaseListener
      */
     public function setupLogging(Event $event)
     {
+        if (!Configure::read('debug')) {
+            return;
+        }
+
         foreach ($this->_getSources() as $connectionName) {
             try {
                 $connection = $this->_getSource($connectionName);


ADmad:
If you don't want any queries to be logged why not simply disable/remove the listener?

asgraf:
I have debugging enabled for white-listed set of IP adressed.
All i want it is `queryLog` key in my json output.

Dumping SQL queries to `queries.log` is unwanted side effect for me.

ADmad:
While you don't want the queries to be logged there might be other that do. So your patch is unacceptable in it's current form. You can update it to allowing configuring the logging behavior while keeping current default behavior.

ADmad:
The listener use `Crud\Log\QueryLogger` to log the queries, which currently isn't configurable. So the proper way to handle this is allow the logger class to be configurable so that you can setup a custom logger which just keeps the logs in memory to be shown in response and not use any persistent logging.

asgraf:
https://book.cakephp.org/3.0/en/orm/database-basics.html#query-logging
CakePHP manual says: 
> Query logging is only intended for debugging/development uses.
> You should never leave query logging on in production as it will negatively impact the performance of your application.

https://crud.readthedocs.io/en/latest/listeners/api-query-log.html
Documentaction of API Query Log says 
> The listener will only append the queryLog key if debug is set to true

> While you don't want the queries to be logged there might be other that do
All im saying is that eating disk space by ApiQueryLogListener (when debug=false) is undocumended behavior and it is unwanted on production enviroments.

asgraf:
https://book.cakephp.org/3.0/en/orm/database-basics.html#query-logging
CakePHP manual says: 
> Query logging is only intended for debugging/development uses.
> You should never leave query logging on in production as it will negatively impact the performance of your application.

https://crud.readthedocs.io/en/latest/listeners/api-query-log.html
Documentaction of API Query Log says: 
> The listener will only append the queryLog key if debug is set to true

All im saying is that logging to files & eating disk space by ApiQueryLogListener (when debug=false) is undocumended behavior and it is unwanted on production enviroments.

> While you don't want the queries to be logged there might be other that do

If someone depends on undocumented features it asks himself for trouble
Logging SQL queries when debugging is disabled is undocumented, unexpected and unintuitive and it shoud not be enabled by default IMHO

asgraf:
> So the proper way to handle this is allow the logger class to be configurable so that you can setup a custom logger which just keeps the logs in memory to be shown in response and not use any persistent logging.

I dont understand why you need to enable logging in setupLogging method then if it wont be used later in beforeRender event in case of debug=false
https://github.com/FriendsOfCake/crud/blob/master/src/Listener/ApiQueryLogListener.php#L77

ADmad:
I believe even with this change, if someone has explicitly enabled query logging through connection  config in `app.php` it should continue working. So I guess it's fine to merge. I'll just wait to see if someone else has any opinion / feedback on this.

Meanwhile I think you are getting queries logged because you have a logger configured with `queriesLog` scope as shown [here](https://github.com/cakephp/app/blob/master/config/app.default.php#L335-L341). So if you just remove that your problem would be solved.

davidyell:
This feels like a custom logger implementation to me.

PR #629, opened 2019-10-14
Update ApiPaginationListener.php

jeanvier:
https://book.cakephp.org/4.0/en/appendices/4-0-migration-guide.html#id1
_"Cake\Controller\Component\PagingComponent now sets paging params info as request attribute instead of request param. Hence you should now use $request->getAttribute('paging') instead of $request->getParam('paging')."_


ADmad:
Fix looks good but related tests also need to be updated.

PR #631, opened 2019-11-30
feat: create Plugin class

josegonzalez:


josegonzalez:
Seems these are unnecessary since Application::addPlugin() takes a string and generates a BasePlugin class.

saeideng:
yes , empty class is optional

Issue #632, opened 2020-01-02
ApiQueryLogListener enables DebugKit query logs


raul338:
Currently ApiQueryLogListener enables queryLogging in **ALL** connections, this includes DebugKit.

If you also have queries log enabled to file, it will fill with DebugKit unnecesary huge queries (panels, history, variables, etc), making it difficult to follow requests/cli queries.

Maybe it would be better to honor the *log* option in connection configuration, or to have a whitelist/blacklist on witch connection to log. Or simply ignore debug_kit logging

PR #638, opened 2020-05-27
Adds option to control which connections to log

Phally:
Had a similar issue as #632 where I needed to exclude a connection.

It was with the Muffin/webservices plugin, which expects a PSR logger, but got the cake query logger.

Also fix a minor unrelated style error.

### DIFF
--- a/tests/TestCase/Action/AddActionTest.php
+++ b/tests/TestCase/Action/AddActionTest.php
@@ -85,7 +85,7 @@ class AddActionTest extends IntegrationTestCase
             'Controller.initialize',
             ['priority' => 11],
             function ($event) {
-                $this->_controller->Flash = $this->getMockBuilder('Cake\Controller\Component\FlashComponent')
+                $this->_controller->Flash = $this->getMockBuilder(\Cake\Controller\Component\FlashComponent::class)
                     ->setMethods(['set'])
                     ->disableOriginalConstructor()
                     ->getMock();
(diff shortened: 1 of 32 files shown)


ADmad:
> It was with the Muffin/webservices plugin, which expects a PSR logger, but got the cake query logger.

In Cake 4 the `Connection` uses PSR logger (`Cake\Database\Log\QueryLog` was updated accordingly) . So are you sure this fix is needed for `cake-4.x` too and not just `master`?

Phally:
Oh excellent! That's great news!

Well this PR wasn't necessarily just a fix for my specific issue. It also works for #632 or just to limit the output when debugging. I see the purpose of this and this Listener for debugging anyways.

Phally:
Changed to draft, so I can fix the `::class` thing.

ADmad:
Yes being able to configure the connections is good. Just didn't want the feature to be added under false pretext :slightly_smiling_face:.

tests/TestCase/Action/AddActionTest.php
```diff
@@ -85,7 +85,7 @@ public function testActionPost()
             'Controller.initialize',
             ['priority' => 11],
             function ($event) {
-                $this->_controller->Flash = $this->getMockBuilder('Cake\Controller\Component\FlashComponent')
+                $this->_controller->Flash = $this->getMockBuilder(\Cake\Controller\Component\FlashComponent::class)
```
ADmad, line 88: `use` statements at the top would have been better. IDEs/Editors can do that for you.

Phally: Not mine apparently (PHPStorm). I mean, it has a similar feature, but will also change it in the docblocks. I don't see a way to exclude those

Phally: Yeah there's no way. It'll also add a `use` for things only mentioned in docblocks. You'll have to do it with your IDE I guess 😄

Phally: I found it!!

PR #639, opened 2020-05-27
Adds option to control which connections to log (backport to v5/master)

### LINKED ISSUE #632: ApiQueryLogListener enables DebugKit query logs

Phally:
Fixes #632

Backport to v5/master.

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -1,7 +1,7 @@
 ---
 linters:
   phpcs:
-    standard: CakePHP
+    standard: CakePHP3
     extensions: 'php,ctp'
     fixer: true
 
--- a/tests/TestCase/Listener/ApiQueryLogListenerTest.php
+++ b/tests/TestCase/Listener/ApiQueryLogListenerTest.php
@@ -37,7 +37,7 @@ class ApiQueryLogListenerTest extends TestCase
     public function testImplementedEvents()
     {
         $Instance = $this
-            ->getMockBuilder('\Crud\Listener\ApiQueryLogListener')
+            ->getMockBuilder(\Crud\Listener\ApiQueryLogListener::class)
             ->disableOriginalConstructor()
             ->setMethods(['_checkRequestType'])
             ->getMock();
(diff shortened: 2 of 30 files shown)


tests/TestCase/Listener/ApiQueryLogListenerTest.php
```diff
@@ -198,6 +198,71 @@ public function testSetupLogging()
...
+     */
+    public function testSetupLoggingConfiguredSources()
+    {
+        $methodName = 'enableQueryLogging';
+        if (version_compare(Configure::version(), '3.7.0RC', '<')) {
+            $methodName = 'logQueries';
+        }
+
+        $DefaultSource = $this
+            ->getMockBuilder('\Cake\Database\Connection')
```
ADmad, line 214: Please `Connection::class` instead of string, same for other classes.

Phally: Haha I just copied it from an existing test case above. Will update both.

ADmad: I you want you can update existing ones too :slightly_smiling_face:

Phally: Yeah that is the plan. Going to look through the plugin and fix them.

Phally:
Okay, I find/replace-ed the string class names per your request 😄

ADmad:
Please add info about this features to the docs at `docs/listeners/api-query-log.rst`

Phally:
> Please add info about this feature to the docs at `docs/listeners/api-query-log.rst`

Oops forgot! I've added it in both.

Phally:
@ADmad Because Stickler does a correction, codesniffer now fails.

How am I supposed to fix this? Stickler will keep doing the correction.

They have a different opinion what alphabetically exactly is.. 😒

.stickler.yml
```diff
@@ -1,7 +1,7 @@
 ---
 linters:
   phpcs:
-    standard: CakePHP
+    standard: CakePHP3
```
ADmad, line 4: It's called `CakePHP3` now? Don't remember such a change being done for core.

Phally: Well that's exactly where I stole it from:

https://github.com/cakephp/cakephp/commit/8a10d4e81ed190d8129e0c1be07dfb93b2732a78

PR #657, opened 2021-01-28
Added example for extra key in ApiListener Action

gringlas:
I thought the documentation could get a bit improved here with an example of how exactly add an extra key to response, as up for today I never really figured it out and even then it took me some trying ;)

### DIFF
--- a/docs/listeners/api.rst
+++ b/docs/listeners/api.rst
@@ -158,21 +158,22 @@ For example you would like to add variable ``my_extra_data`` you have to add it 
 
 .. code-block:: php
 
-class MyAction extends BaseAction 
-{
-
-  protected $_defaultConfig = [
-    'serialize' => [
-      'my_extra_data'
-    ],
-    // other stuff
-  ];
-  
-  protected function handle()
+  <?php
+  class MyAction extends BaseAction 
   {
-    // do other stuff here
-    $this->_controller()->set('my_extra_data', 'This is my extra data I want to add to response');
-  }
+
+    protected $_defaultConfig = [
+      'serialize' => [
+        'my_extra_data'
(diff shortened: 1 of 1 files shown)


docs/listeners/api.rst
```diff
@@ -154,6 +154,35 @@ Response format
 
 The default response format for both XML and JSON has two root keys, ``success`` and ``data``. It's possible to add
 your own root keys simply by using ``_serialize`` on the view var.
+For example you would like to add variable ``my_extra_data`` you have to add it to the configuration of the action like, this:
+
+.. code-block:: php
+
+  <?php
+  class MyAction extends BaseAction
```
ADmad, line 162: Instead of an example using a custom action class, an example which uses `$this->Crud->action()->setConfig()` would be more suitable.

gringlas: You mean like, so?

```
class SomeController extends AppController {

public function initialize() {
      $this->Crud->setConfig([
            'serialize' => [
                'my_extra_data'
            ],
            'api' => [
                'success' => [
                    'data' => [
                        'entity' => [],
                        'my_extra_data'
                    ]
                ]
            ]
        ]);
}
```

ADmad: Yes, though you missed the `->action()` in the call chain and it should be done in `beforeFilter()` or controller's action method.

irongomme: Personnaly, I'm using it like this : 

```
$this->set(compact('aggregation'));
$this->Crud->action()->setConfig('serialize.aggregation', 'aggregation');
```

gringlas:
Why did those checks fail?

PR #664, opened 2021-04-21
Overwrite isAction

Zwartpet:
In https://github.com/FriendsOfCake/crud/pull/645 the isAction overwrite was removed.
If I use the AuthComponent in combination with the CrudComponent the authentication is not triggered because isAction is false.

Is there a different way to make this work or is the AuthComponent not longer supported in combination with CrudComponent?

### DIFF
--- a/src/Controller/ControllerTrait.php
+++ b/src/Controller/ControllerTrait.php
@@ -52,6 +52,27 @@ trait ControllerTrait
     }
 
     /**
+     * Return true for a mapped action so that AuthComponent doesn't skip
+     * authentication / authorization for that action.
+     *
+     * @param string $action Action name
+     * @return bool True is action is mapped and enabled.
+     */
+    public function isAction(string $action): bool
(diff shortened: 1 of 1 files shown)


ADmad:
Overriding `Controller::isAction()` was no longer necessary for the working of the plugin in Cake 4.x and hence it was removed. Also `AuthComponent` has been deprecated in Cake 4.x.

You can just override `isAction()` in your `AppController` class if you are still using the `AuthComponent`.

AdrianHL:
I have spent nearly a day finding this problem after upgrading an application from 3 to 4. I'd be nice to see this type of breaking changes, specially if these are around authentication/authorization, in some kind of upgrade guide/notes. You cannot assume that all applications will also replace the AuthComponent in the same step. That component is deprecated and to be replaced in 5.X so there is no reason to require, or if so, again, make it clear some how in an upgrade guide.

PR #680, opened 2022-10-27
Use CakeException in CrudException

stefanozoffoli:
Hi all,
I got an error with composer autoload:
error: [Error] Class 'Cake\Core\Exception\Exception' not found in /srv/www/vendor/friendsofcake/crud/src/Error/Exception/CrudException.php on line 9

I fixed the problem by changing the class reference like this. I hope it's the right way to fix this.

Thank you


ADmad:
Well then you need to figure out how to get the same Cake and pluign's version on your prod server as you have locally.

PR #706, opened 2023-10-24
Ability to replace primaryKey field via config

asgraf:
Example use case:
```php
$this->Crud->action()->setConfig('primaryKey', 'uuid');
```
this change makes `FriendsOfCake/crud` plugin compatible with `dereuromark/cakephp-expose`


ADmad:
- Test(s) required for the change.
- The `primaryKey` config should be added to `$_defaultConfig` of all actions that use the `FindMethodTrait::_findRecord()` method.
- Docs need to be updated to provide info about this new option.

PR #707, opened 2023-10-24
Ability to replace primaryKey field via config (6.x)

asgraf:
Example use case:
```php
$this->Crud->action()->setConfig('primaryKey', 'uuid');
```
this change makes `FriendsOfCake/crud` plugin compatible with `dereuromark/cakephp-expose`

### DIFF
--- a/src/Traits/FindMethodTrait.php
+++ b/src/Traits/FindMethodTrait.php
@@ -61,11 +61,12 @@ trait FindMethodTrait
 
         [$finder, $options] = $this->_extractFinder();
         $query = $repository->find($finder, $options);
+        $primaryKey = $this->getConfig('primaryKey', $repository->getPrimaryKey());
         /**
          * @psalm-suppress PossiblyInvalidArgument
          * @psalm-suppress InvalidArrayOffset
          */
-        $query->where([current($query->aliasField($repository->getPrimaryKey())) => $id]);
+        $query->where([current($query->aliasField($primaryKey)) => $id]);
 
         $subject->set([
             'repository' => $repository,


ravage84:
Wouldn't it make more sense to change this in the table itself?

Or why would one change this for the CRUD component but not in the table (thus for the whole application)?

A primary key normally doesn't get changed... 🤷

asgraf:
> A primary key normally doesn't get changed... 🤷

Yes, normally it does not (see [dereuromark/cakephp-expose](https://github.com/dereuromark/cakephp-expose/) for details)

> why would one change this for the CRUD component but not in the table (thus for the whole application)?

`$this->Crud->action()->setConfig('primaryKey', 'uuid');` applies this change only for request generated by Crud action

With cake-expose you can have urls like:
[/view/SAftQtJqYjAs3XRMsWgRfk](https://sandbox4.dereuromark.de/sandbox/expose-examples/view/SAftQtJqYjAs3XRMsWgRfk) instead of `/view/3` (url contains short-uuid instead numeric id)




For url like `/view/SAftQtJqYjAs3XRMsWgRfk` Crud would try to generate SQL like this:
```sql
SELECT [fields list] FROM articles as Articles WHERE Articles.id = 'SAftQtJqYjAs3XRMsWgRfk'
```
This would fail because `id` is `autoincrement int` field
After this patch with `$this->Crud->action()->setConfig('primaryKey', 'uuid');` and `dereuromark/cakephp-expose` Crud will generate valid SQL as follows:
```sql
SELECT [fields list] FROM articles as Articles WHERE Articles.uuid = 'efbfbdef-bfbd-7d56-efbf-bdefbfbd41ef'
```
(`SAftQtJqYjAs3XRMsWgRfk` is short version of `efbfbdef-bfbd-7d56-efbf-bdefbfbd41ef` because `dereuromark/cakephp-expose` uses [keiko/uuid-shortener](https://github.com/mgrajcarek/uuid-shortener))

> Wouldn't it make more sense to change this in the table itself?

Changing primary key at table level would break all request with `$query->contain()` in whole app (foreign keys are numeric ids not uuids). See `dereuromark/cakephp-expose` for details

davidyell:
Seems like a test is needed to prove the functionality

dereuromark:
@asgraf Are you still adding the test? Then this could get merged.

PR #711, opened 2024-01-18
Use primary key instead of binding key in case of BelongsToMany association

ravage84:
Belongs to many associations have a junction table or through model:

https://book.cakephp.org/4/en/orm/associations.html#belongstomany-associations

A field named like the binding key does not necessarily exist on the target table but only on the junction table.

This change is necessary to prevent the exception introduced by the backport of https://github.com/cakephp/cakephp/pull/17340 to be thrown in such cases.

https://github.com/cakephp/collection/commit/38971692fc079357a93e9b0b26faa5c606073fc2#diff-81970ebd43ab052712b24c8315dfd428fea23a3dcea21316e27d9ce9a2551b83R596-R599

I encountered this after upgrading an application from CakePHP 4.4.x. to 4.5.x.

The reason why this has worked before 4.5 is mentioned at https://github.com/cakephp/cakephp/pull/17340#issuecomment-1898926391.

This change is likely also necessary for the CRUD branch for CakePHP 5.x.

### DIFF
--- a/src/Listener/RelatedModelsListener.php
+++ b/src/Listener/RelatedModelsListener.php
@@ -118,6 +118,12 @@ class RelatedModelsListener extends BaseListener
      */
     protected function _findOptions(Association $association): array
     {
+        if ($association instanceof Association\BelongsToMany) {
+            return [
+                'keyField' => $association->getPrimaryKey(),
+            ];
+        }
+
         return [
             'keyField' => $association->getBindingKey(),
         ];
--- a/src/Listener/RelatedModelsListener.php
+++ b/src/Listener/RelatedModelsListener.php
@@ -6,6 +6,7 @@ namespace Crud\Listener;
 use Cake\Datasource\EntityInterface;
 use Cake\Event\EventInterface;
 use Cake\ORM\Association;
+use Cake\ORM\Association\BelongsToMany;
 use Cake\Utility\Inflector;
 use RuntimeException;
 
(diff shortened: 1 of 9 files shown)


src/Listener/RelatedModelsListener.php
```diff
@@ -118,6 +118,12 @@ public function publishRelatedModels(?string $action = null, ?EntityInterface $e
...
     {
+        if ($association instanceof Association\BelongsToMany) {
+            return [
+                'keyField' => $association->getPrimaryKey(),
+            ];
+        }
+
         return [
             'keyField' => $association->getBindingKey(),
         ];
```
ravage84, line 129: Alternatively, the code could look like this:

```suggestion
        $keyField = $association->getBindingKey();
        if ($association instanceof Association\BelongsToMany) {
            $keyField' = $association->getPrimaryKey();
        }

        return [
            'keyField' =>$keyField,
        ];
```

src/Listener/RelatedModelsListener.php
```diff
@@ -118,6 +118,12 @@ public function publishRelatedModels(?string $action = null, ?EntityInterface $e
      */
     protected function _findOptions(Association $association): array
     {
+        if ($association instanceof Association\BelongsToMany) {
```
ADmad, line 121: Please add a `use` statement for the class.

ADmad (commented):
Test case required.

BTW the docs say it's the column of the "current" table, not junction table.

> bindingKey: The name of the column in the current table, that will be used for matching the foreignKey. Defaults to the primary key.

P.S. The `master` branch is for CakePHP 5, I can backport it later.

dereuromark:
ping @ravage84 
Is this still sth that should be done?

ravage84:
I believe so, yes.

PR #712, opened 2024-03-28
Fix query logger

JustinRuiter:
Really simple PR, to restore the original workings of the query logger. In previous versions of crud, the query logger would not only store the query, but also the entire context, like the query execution time and rows affected. Right now, a cast to string means only the query is stored in the query log. This PR attempts to restore the original functionality, by using the jsonSerialize method for instances of LoggedQuery.

### DIFF
--- a/src/Controller/Component/CrudComponent.php
+++ b/src/Controller/Component/CrudComponent.php
@@ -233,7 +233,7 @@ class CrudComponent extends Component
     {
         $this->_loadListeners();
 
-        $this->_action = $controllerAction ?: $this->_action;
+        $this->_action = $controllerAction ?? $this->_action;
 
         $action = $this->_action;
         if (empty($args)) {
@@ -275,7 +275,7 @@ class CrudComponent extends Component
      */
     public function action(?string $name = null): BaseAction
     {
-        if (empty($name)) {
+        if ($name === null) {
             $name = $this->_action;
         }
 
@@ -425,7 +425,7 @@ class CrudComponent extends Component
      */
     public function isActionMapped(?string $action = null): bool
     {
-        if (empty($action)) {
+        if ($action === null) {
             $action = $this->_action;
         }
 
@@ -448,7 +448,7 @@ class CrudComponent extends Component
     public function on(array|string $events, callable $callback, array $options = []): void
     {
         foreach ((array)$events as $event) {
-            if (!strpos($event, '.')) {
+            if (!str_contains($event, '.')) {
                 $event = $this->_config['eventPrefix'] . '.' . $event;
             }
 
--- a/src/Log/QueryLogger.php
+++ b/src/Log/QueryLogger.php
@@ -3,6 +3,7 @@ declare(strict_types=1);
 
 namespace Crud\Log;
 
+use Cake\Database\Log\LoggedQuery;
 use Cake\Database\Log\QueryLogger as CakeQueryLogger;
 use Stringable;
 
@@ -30,7 +31,11 @@ class QueryLogger extends CakeQueryLogger
      */
     public function log($level, string|Stringable $message, array $context = []): void
     {
-        $this->_logs[] = (string)$context['query'];
+        if ($context['query'] instanceof LoggedQuery) {
+            $this->_logs[] = $context['query']->jsonSerialize();
+        } else {
+            $this->_logs[] = (string)$context['query'];
+        }
 
         parent::log($level, $message, $context);
     }
--- a/src/Log/QueryLogger.php
+++ b/src/Log/QueryLogger.php
@@ -32,6 +32,7 @@ class QueryLogger extends CakeQueryLogger
     public function log($level, string|Stringable $message, array $context = []): void
     {
         if ($context['query'] instanceof LoggedQuery) {
+            /** @psalm-suppress InternalMethod **/
             $this->_logs[] = $context['query']->jsonSerialize();
         } else {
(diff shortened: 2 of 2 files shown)


src/Controller/Component/CrudComponent.php
```diff
@@ -275,7 +275,7 @@ public function execute(?string $controllerAction = null, array $args = []): Res
      */
     public function action(?string $name = null): BaseAction
     {
-        if (empty($name)) {
+        if ($name === null) {
```
davidyell, line 278: Not sure what the point of this change is

src/Controller/Component/CrudComponent.php
```diff
@@ -448,7 +448,7 @@ public function isActionMapped(?string $action = null): bool
     public function on(array|string $events, callable $callback, array $options = []): void
     {
         foreach ((array)$events as $event) {
-            if (!strpos($event, '.')) {
+            if (!str_contains($event, '.')) {
```
davidyell, line 451: Similar here, is this more performant? Otherwise seems a bit pointless

src/Controller/Component/CrudComponent.php
JustinRuiter: These are changes of lines that psalm was complaining about, to make the pipeline pass. If not passing the pipeline is not a problem, the Existing psalm errors fixed commit can be removed.

src/Controller/Component/CrudComponent.php
JustinRuiter: Same as the other comment

src/Controller/Component/CrudComponent.php
davidyell: Looks to me like the coding standards passed. The failures look like test coverage

JustinRuiter: Yes, now they pass, but without these changes, coding standards do not pass. I assume this was already in master, as I branched from master yesterday.

src/Log/QueryLogger.php
```diff
@@ -30,7 +31,12 @@ public function getLogs(): array
      */
     public function log($level, string|Stringable $message, array $context = []): void
     {
-        $this->_logs[] = (string)$context['query'];
+        if ($context['query'] instanceof LoggedQuery) {
+            /** @psalm-suppress InternalMethod **/
+            $this->_logs[] = $context['query']->jsonSerialize();
```
ADmad, line 36: Looking at Cake 3.x the casting `LoggedQuery` to string gave the format `"duration={$this->took} rows={$this->numRows} {$this->query}"` as seen [here](https://github.com/cakephp/cakephp/blob/3.x/src/Database/Log/LoggedQuery.php#L67). So if you want to retain old format that's what needs to be done here instead of using the array provided by `jsonSerialized()`.

You can use the `LoggedQuery::getContext()` method to get the details and construct a string with the above format.

JustinRuiter: Logging the string message in that format is handled by the parent call to the cake QueryLogger::log(). getLogs() always returned an array of dictionaries in the format that jsonSerialize() returns.

src/Controller/Component/CrudComponent.php
JustinRuiter: If that is a good and clear explanation, can you please close the conversation?

src/Log/QueryLogger.php
JustinRuiter: Does this seem right to you? @ADmad

ADmad: In your initial comment you said you want to restore the original workings of the query logger. The changes I suggested to achieve the format used in Cake 3 compatible version of the plugin (since Cake 4 casing `LoggedQuery` to string just return the SQL query). `getLogs()` has always returned an array of string, not an array or arrays.

JustinRuiter: It seems to me that the query logger did return an array of LoggedQueries actually, please refer to commit 9aa30db8aa68b4da453f5fc9b062732c7d7fc93d , where the _log method appends the LoggedQuery object to _logs[].

ADmad: Hmm.. I checked again and indeed that seems to be the case. So I guess all we need to do is remove the string casting and update the docblock of `$_logs` property to indicate it's an array of `LoggedQuery`.

BTW where are you seeing/using the array returned by `getLogs()`?  `ExceptionRenderer` and `ApiQueryLogListener` sets the array as a view var. When the vars are serialized for JSON/XML response the instances would be cast to string and you will still end up with just the SQL query and before. So what does this change achieve?

JustinRuiter: We are using a custom `ApiQueryLogListener` that also returns (for every connection) the array of `LoggedQuery` objects. When accessing the page, a `queryLog` key is appended to the output, containing the query that was executed, the number of rows affected, parameters, how long the query took and any errors. It seems to me that this is the result of json serializing using the `jsonSerialize()` method in `LoggedQuery`.

query log

JustinRuiter: Okay, implemented that in the latest comit

src/Log/QueryLogger.php
```diff
@@ -3,6 +3,7 @@
 
 namespace Crud\Log;
 
+use Cake\Database\Log\LoggedQuery;
```
ADmad, line 6:
```suggestion
```

PR #743, opened 2024-10-02
chore: drop extra dependencies

josegonzalez:
We only really care about a handful of dependencies, and the rest can be auto-installed by rtd.


ADmad:
```
Extension error:
Could not import extension sphinxcontrib.phpdomain (exception: No module named 'sphinxcontrib.phpdomain')
```

PR #753, opened 2025-01-27
Fix argument type for RedirectTrait methods

ADmad:


davidyell (approved):
Seems sensible.

PHPStan has an errors baseline which might be neater, but I'm not sure if Psalm has a similar feature or not.

PR #761, opened 2025-07-05
Cleanup.

dereuromark:
as entity() directly patches.
Also added missing tests for entity() along with it, and fixed some boilerplate plugin stuff similar to other FOC plugins.

### DIFF
--- a/src/Core/Object.php
+++ b/src/Core/Object.php
@@ -1,7 +1,0 @@
-<?php
-declare(strict_types=1);
-
-// @deprecated Add backwards compat alias. "Object" is protected keyword in PHP 7.2
-if (PHP_VERSION_ID < 70200) {
-    class_alias('Crud\Core\BaseObject', 'Crud\Core\Object');
-}


src/Controller/Component/CrudComponent.php
```diff
@@ -660,6 +660,16 @@ public function entity(array $data = []): EntityInterface
         return $this->model()->newEntity($data);
     }
 
+    /**
+     * Returns new empty entity.
+     *
+     * @return \Cake\Datasource\EntityInterface
+     */
+    public function emptyEntity(): EntityInterface
```
ADmad, line 668: I don't understand why do you need this method?

dereuromark: I thought this completed/complemented the existing entity($data) method when you have to do one action manually, and therefore need sometimes the former or latter.
One could also manually call the table, but it seems this one has the internal context and knows the correct tables right away.

ADmad: Nothing wrong with directly using the table instance.

I want to decrease the API surface rather than increasing. Just something I haven't got around to. `CrudComponen::entity()` isn't used within the plugin and should be deprecated. `CrudComponent::useModel()` too isn't needed. If one needs a non default/conventional model they can just set `Controller::$defaultTable` and everything under Crud will use that table.

dereuromark: I see

ADmad:
The `src/Core/Object.php` file can be dropped. That will fix the stan error.

dereuromark:
Yeah I was suprised it was still in there as this shouldnt be since the 7.2 days (< Cake5)

PR #762, opened 2025-07-05
Sync the add/edit actions. PHPUnit cleanup.

dereuromark:
Edit seemed to have missed the stopping on the callback,
now add/edit are synced in behavior.

### DIFF
--- a/src/Action/AddAction.php
+++ b/src/Action/AddAction.php
@@ -116,8 +116,9 @@ class AddAction extends BaseAction
      */
     protected function _post(): ?Response
     {
+        $entity = $this->_entity($this->_request()->getData(), $this->saveOptions());
         $subject = $this->_subject([
-            'entity' => $this->_entity($this->_request()->getData(), $this->saveOptions()),
+            'entity' => $entity,
             'saveMethod' => $this->saveMethod(),
             'saveOptions' => $this->saveOptions(),
         ]);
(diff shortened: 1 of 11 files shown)


dereuromark:
I dont know what the failing test has to do with it.

dereuromark:
Once my PRs are through, maybe I can showcase it a bit in the sandbox, with some live example code to check out
https://github.com/dereuromark/cakephp-sandbox/pull/94

src/Action/AddAction.php
```diff
@@ -127,9 +128,8 @@ protected function _post(): ?Response
             return $this->_stopped($subject);
         }
 
-        $saveCallback = [$this->_model(), $subject->saveMethod];
         /** @phpstan-ignore argument.type */
-        if (call_user_func($saveCallback, $subject->entity, $subject->saveOptions)) {
+        if (call_user_func([$this->_model(), $this->saveMethod()], $entity, $this->saveOptions())) {
```
ADmad, line 132: This is just wrong. The whole point of setting the entity, saveMethod and saveOptions to the subject is to allow using customizing them using the `Crud.beforeSave` event. Hence those values should be used from the subject after the event execution.

dereuromark: Then edit must be wrong, as those are doing different ways of handling this.

ADmad:
Changes to `EditAction` are causing test failure.

dereuromark:
https://github.com/FriendsOfCake/crud/pull/766

ADmad:
Closing in favor of #766

PR #763, opened 2025-07-05
Poc adit

dereuromark:
Just a try at adit logic
Actually works quite easily
But I think the testing and also template links are bit more tricky I assume


ADmad:
Using Crud your controllers are already very lean, so no need to go overboard and merge the two actions too.

Issue #764, opened 2025-07-06
Docs Query


dereuromark:
I saw in the docs:

    public function _beforeFind(\Cake\Event\EventInterface $event, \Cake\ORM\Query $query)

should this be SelectQuery?

ADmad:
Yes

dereuromark:
I guess changing existing code isnt possible though

    abstract protected function _bulk(Query $query): bool;

etc

ADmad:
That type is fine, for bulk action it can be a query of any type, not just `SelectQuery`:

https://github.com/FriendsOfCake/crud/blob/60773c205549b134c76c7469e1346171ea114f32/src/Action/Bulk/SetValueAction.php#L64-L66

PR #765, opened 2025-07-06
Fix up SelectQuery examples.

dereuromark:
Resolves https://github.com/FriendsOfCake/crud/issues/764

### DIFF
--- a/docs/listeners/custom.rst
+++ b/docs/listeners/custom.rst
@@ -23,9 +23,9 @@ the ``Crud.beforeFind`` event to the ``_beforeFind()`` method in the controller.
         ];
     }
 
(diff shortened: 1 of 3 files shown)


tests/test_app/src/Model/Endpoint/SearchEndpoint.php
```diff
@@ -1,14 +0,0 @@
-<?php
-declare(strict_types=1);
-
-namespace Crud\Test\App\Model\Endpoint;
-
-use Muffin\Webservice\Model\Endpoint;
```
ADmad, line 6: Revert this deleting, it's for the webservice plugin which might be upgrade for Cake5 someday.

PR #766, opened 2025-07-06
Sync add/edit actions.

ADmad:
Handle "beforeSave" event being stopped for EditAction.


ADmad:
@dereuromark This is technically a new feature for the `EditAction` so I am going to put it in the next minor (have started a `next` branch).

PR #767, opened 2025-07-06
Deprecate methods

ADmad:


ADmad:
Aargh... @dereuromark Any idea how to get the build passing for php 8.1? I updated the phpunit config as per the changes you did here https://github.com/cakephp/cakephp/commit/9b973bb4a497c4ab4248729b9c4936fb0d248c4f

dereuromark:
Not sure what it is, but

    "cakephp/cakephp": "^5.2.3"
    
seems to fix it, so sth we changed in that version regarding deprecations.
https://github.com/cakephp/cakephp/compare/5.2.2...5.2.3

maybe is the errorlevel overwrite in there, so plugins might need their own updated version of it to work across the dependency spectrum.

